PR #4, opened 2016-06-06
pyqt5 compatibility

tobiasgehring:
requires development version of pyqtgraph for pyqt5. It is still compatible with pyqt4. If both are installed on a system pyqt4 will be used. To use pyqt5 you have to change the variable libOrder in pyqtgraph/Qt.py

### DIFF
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -27,7 +27,7 @@ __version__ = '0.1'
 import os
 import sys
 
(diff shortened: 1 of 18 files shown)


drogenlied:
Thanks for the work!
There was one conflict with the curent master branch but that was easier to resolve than asking for it to be fixed.

tobiasgehring:
Thanks. github told me that there were conflicts because I was working
on an older version, but I hoped that it would be simple to fix them.
Thanks for doing that.

PR #6, opened 2016-06-29
fixed qt5 message handler which has an additional argument

tobiasgehring:
didn't notice that when I introduced qt5 support


Neverhorst:
Nice... actually I already encountered errors related to that while using the software. Thanks for fixing.

PR #8, opened 2016-07-11
changed name of default_* callback functions to on_*.

tobiasgehring:
These methods should be overridden in a descendant class instead of passing new callback functions via the constructor. The old behavior is still possible and besides of backward compatibility it allows to define new callback functions not defined in the base class. The latter is what it should be used for in the future.

### DIFF
--- a/core/base.py
+++ b/core/base.py
@@ -65,8 +65,8 @@ class Base(QtCore.QObject, Fysom):
         QtCore.QObject.__init__(self)
 
         default_callbacks = {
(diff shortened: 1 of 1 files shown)


jochen-scheuer:
hey guys,

I have an appointment at 2pm in neu-ulm at the dentist. could we please postpone it to 5pm?

If not possible I can not join.

See you later,
Jochen

drogenlied:
Yeah, this is definitely the right place to post that.

PR #9, opened 2016-07-26
Logging

tobiasgehring:
new logging facility using the logging module of python


drogenlied:
If you are removing logMsg from Base there should be no logMsg call left anywhere in all of the code.

tobiasgehring:
seems like some where introduced after I removed all of them and so they
slipped. Should be fixed now.

drogenlied:
Still some left in hardware/spincore/pulse_blaster_esrpro.py

tobiasgehring:
in this file logMsg is defined itself and simply prints the message. No need to change that from my side.

On Tue, Jul 26, 2016 at 2:37 PM, Jan Binder < notifications@github.com [notifications@github.com] > wrote:
Still some left in hardware/spincore/pulse_blaster_esrpro.py

a-stark:
that is true, I will change that after the merge, since that module was tested independently from the qudi, and the logMsg was overwritten in that file for testing reasons.

drogenlied:
Oh, that is right. I will bother Alex with that.

PR #10, opened 2016-07-27
Support for cooperative multiple inheritance

tobiasgehring:
This patch enables support for cooperative multiple inheritance for modules with Qt5. It furthermore reduces the complexity of the constructor in modules:

instead of
def __init__(self, manager, name, config, **kwargs):
    super().__init__(manager, name ,config, **kwargs)

you can write
def __init__(self, config, **kwargs):
    super().__init__(config=config, **kwargs)
if config is used in the constructor or simply
def __init__(self, **kwargs):
    super().__init__(**kwargs)
if it is not.

This patch furthermore removes all callback definitions in the modules and directly uses on_activate and on_deactivate as introduced earlier.

The advantage of cooperative multiple inheritance with Qt5 comes into play in GUI modules where composition is no longer needed but modules can be derived from QMainWindow and Base at the same time.


drogenlied:
Loading all modules from the example configuration produces quite a few exceptions.
See [qudi.log.txt](https://github.com/Ulm-IQO/qudi/files/385952/qudi.log.txt)

tobiasgehring:
fixed. Some mistakes when I converted the constructors.

PR #12, opened 2016-08-01
Module bugfixes

tobiasgehring:
More parameter checking to prevent exception when deactivating not loaded module + some more checking and testing if module is active before deactivating it (e.g. at reload) to prevent unnecessary error messages.

### DIFF
--- a/core/manager.py
+++ b/core/manager.py
@@ -746,10 +746,12 @@ class Manager(QtCore.QObject):
             except:
                 logger.exception('Error while loading {0} module: '
                         '{1}'.format(base, key))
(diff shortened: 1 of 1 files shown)


drogenlied:
I (and other people) would disagree that the decorators are unnecessary, we probably should be using more of them in other places:
https://stackoverflow.com/questions/14421897/is-the-pyside-slot-decorator-necessary

Tho other changes are good.

tobiasgehring:
I actually didn't know that. Learned something for today, thank you!

I reverted the commit.

PR #24, opened 2016-09-08
Mapper

tobiasgehring:
this is a new core module to establish mappings between a displaying widget in a gui module and a property (or getter and setter functions) in e.g. a logic module.

The purpose of it is to enable automatic updates in both directions. If a user changes the value in a widget, the property is automatically updated. If a user changes the property programatically, the GUI is updated. It seems to me that especially the latter is not implemented in all modules, e.g. the counter module where the count frequency can be changed in the logic module, but the GUI isn't updated.

I tried to implement this as general as possible and so far it covers all my own use cases, but it could be that something is still missing. In case, I'm happy to add such use cases.

Please see the documentation of the Mapper class for an example.

### DIFF
--- a/core/mapper.py
+++ b/core/mapper.py
@@ -118,6 +118,26 @@ class Mapper():
         self._submit_policy = SUBMIT_POLICY_AUTO
         self._mappings = {}
 
+    def _get_property_from_widget(self, widget):
+        """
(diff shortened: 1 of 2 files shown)


drogenlied:
Nice, that should make it way easier for people to keep the GUI an as thin layer as possible.

PR #25, opened 2016-10-14
Pulsed with queued connections

Neverhorst:
This branch was for reworking the communication between logic modules and logic and GUI module for pulsed measurements.
The pulsed_maingui is now only connecting to a pulsed_master_logic module. From there on the entire communication with other modules (pulsed_measurement_logic and sequence_generator_logic) is handled via signals instead of direct method calls. For each input/parameter setting from the GUI/master_logic a "set_something" signal is issued to the corresponding logic. After the settings have been taken into account the logic emits a "something_updated" signal upwards to the GUI and the display gets updated accordingly. This has the advantage that threads are properly decoupled and the GUI gets updated even if some external script is manipulating the logic directly.
Complete overhaul of the GUI ui-files and python module.
The only missing part compared to the master branch GUI is the sequence editor and the second plot in the pulse analysis tab. But both functionalities did not work in 90% of the cases before so they need to be reworked anyway.
Over all pulsed measurements run much more stable than before and also some minor features have been added. (FastComtec record length setting, alternating sequences, laser ignore capability, resume old measurement capability (not in GUI yet))


drogenlied:
Software meeting says yes.

PR #26, opened 2016-10-16
slow_counter_interface: use ABC abstract class instead of exceptions

tobiasgehring:
This pull request is a demonstration of how ABC abstract classes can be used instead of the custom exception. Has somebody experience with pyside? This patch should break pyside since I guess pyqtWrapperType is not available. The question is what we have to do to make it work.

It seems that I cannot install pyside using pip with python 3.5...


tobiasgehring:
I think I fixed it for pyside, but it's still untested.

tobiasgehring:
I converted most of the interfaces. I haven't converted pulser_interface and microwave_interface because they define methods that seem not to be used in the implementations.

drogenlied:
PySide does not work, but not because of these changes, as it causes some error in the logger:

```
Specified Qt API: pyside
Used Qt API: PySide
  File "/usr/lib/python3.5/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/jan/qo/qudi-tobias/core/__main__.py", line 59, in <module>
    logger.info('Loading Qudi...')
  File "/usr/lib/python3.5/logging/__init__.py", line 1280, in info
    self._log(INFO, msg, args, **kwargs)
  File "/usr/lib/python3.5/logging/__init__.py", line 1416, in _log
    self.handle(record)
  File "/usr/lib/python3.5/logging/__init__.py", line 1426, in handle
    self.callHandlers(record)
  File "/usr/lib/python3.5/logging/__init__.py", line 1488, in callHandlers
    hdlr.handle(record)
  File "/usr/lib/python3.5/logging/__init__.py", line 856, in handle
    self.emit(record)
  File "/home/jan/qo/qudi-tobias/core/logger.py", line 112, in emit
    self.sigLoggedMessage.emit(record)
TypeError: emit() takes 2 positional arguments but 3 were given
Unexpected return value 1. Exiting.
```
Does pyqt5 work for you? I get some error with the garbage collector.

tobiasgehring:
interesting. Not sure what's the problem there and where the third
argument comes from...

pyqt5 works fine for me, I don't get any error messages. Qt 5.6, python
3.5. What error do you get?

drogenlied:
I get this mysterious error:
```
Specified Qt API: pyqt5
Used Qt API: PyQt5
Loading Qudi...
QObject::connect: Cannot connect QTimer::timeout() to (null)::check()
  File "/usr/lib/python3.5/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/jan/qo/qudi-tobias/core/__main__.py", line 134, in <module>
    gc = GarbageCollector(interval=1.0, debug=False)
  File "/home/jan/qo/qudi-tobias/core/garbage_collector.py", line 42, in __init__
    self.timer.timeout.connect(self.check)
TypeError: connect() failed between timeout() and check()
Unexpected return value 1. Exiting.
```
First I thought that was because of an old pyqtgraph , but installing the one from git dit not change anything.
I have Qt 5.6.1, Python 3.5.1.

drogenlied:
However, this does not happen in the environment tested with TravisCI:
https://travis-ci.org/Ulm-IQO/qudi/builds/168729924

drogenlied:
I tested this a little bit and there were no errors on Windows/Anaconda env.

PR #27, opened 2016-10-16
fixed syntax error introduced in commit fdfc60784c9fac8bfc7c0e6adde8f…

tobiasgehring:
…e32aa671225

please always test changes you make as good as you can to avoid breaking things


drogenlied:
There is no way to test, no unit tests, no modules excercising the code, no __main__ inside the file acting as a quick sanity check.

If you want tests for unused code, please provide them.

PR #32, opened 2016-11-24
Odmr abc

drogenlied:
Convert microwave interface to abstract base class and convert some string literals to enums.

Please test and merge, I do not have a setup available for testing right now.


Neverhorst:
Briefly tested common tasks on my Setup (Setup9) and after a minor bugfix (commit d10c31c) it worked.
It would be good if we had someone else test it on his setup.

PR #33, opened 2016-11-24
Nicard retval

drogenlied:
Ensure that every function communicates success/failure by return value and not by throwing an exception.

Please test and merge.
I do not have a setup for testing right now.


Neverhorst:
Briefly tested common tasks on my Setup (Setup9) and after a minor bugfix (commit 86eafa7) it worked.
It would be good if we had someone else test it on his setup.

PR #156, opened 2017-02-25
Retire out connectors

tobiasgehring:
This pull request removes out connectors.

I made the experience that the whole connector business is very complicated to explain to a novice and very difficult for them to understand the concept. Thus, a simplification will help tremendously. All currently used functionality can be implemented without out connectors and type checking could still be implemented. I have discussed the proposal of removing out connectors with Alexander Stark.

With the removal of out connectors, a connector is connected to any another module. The connector itself is defined - as it was before - by a name and a type. The module that is connected to the connector has to have that type, i.e. it has to implement a certain interface (either defined by an abc abstractclass or a regular class). Due to multiple inheritance a module can implement multiple interfaces. Type checking is not implemented yet and is left for another pull request (type checking was never implemented in qudi before, so no regression here).

Main Changes:
- removal of out connectors
- renamed "_in" dictionary in modules with "_connectors" (though "_in" still works right now and can be deprecated later)
- renamed "get_in_connector" to "get_connector" (no legacy method right now)
- old config files still work and just produce a warning. The second part in the definition of the connectee after the "." is ignored. New style configuration files simply point to a suitable module by the name of the inheritance defined in the config file.

### DIFF
--- a/config/example/default.cfg
+++ b/config/example/default.cfg
@@ -30,7 +30,7 @@ hardware:
         module.Class: 'confocal_scanner_dummy.ConfocalScannerDummy'
         clock_frequency: 100
         connect:
-            fitlogic: 'fitlogic.fitlogic'
+            fitlogic: 'fitlogic'
 
     mydummycounter:
         module.Class: 'slow_counter_dummy.SlowCounterDummy'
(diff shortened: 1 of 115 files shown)


drogenlied:
I'm OK with this.

Originally, the plan was to support multiple channels per device through different connectors, but that never materialised, so now we can get rid of the complexity.

We just need a bit of time to notify people of this change and I need to update the config editor prototype.

PR #159, opened 2017-03-13
Savelogic improvement

Neverhorst:
Changed the savelogic from saving each entry of a list/array explicitly element-wise to something more efficient. Now the input arrays get converted to numpy.ndarrays and are saved by the numpy.savetxt.
Saving of the text files is now much faster (matplotlib plots still slow as hell (txt-saving: 0.2s, matplotlib: 6s)... but that's another story).
Also refined the definition of the save_data method to be more explicit with respect to what it expects from the logic programmer. You can now either pass (multiple) 1D arrays or a single 2D array (most efficient) but no mixture of both. Unequal 1D array length is handled in that way that the shorter arrays will be padded with "nan". This can lead to problems when using integer type arrays since for this type "nan" is not defined.
Tried to adapt all measurement logics to the changes and so far encountered no problems. Please test and merge.

### DIFF
--- a/gui/confocal/confocalgui.py
+++ b/gui/confocal/confocalgui.py
@@ -234,7 +234,6 @@ class ConfocalGui(GUIBase):
         self._scanning_logic = self.get_in_connector('confocallogic1')
         self._save_logic = self.get_in_connector('savelogic')
         self._optimizer_logic = self.get_in_connector('optimizerlogic1')
-        self._save_logic = self.get_in_connector('savelogic')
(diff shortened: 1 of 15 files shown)


drogenlied:
Looks OK so far.

latchr:
Also looks good to me.  I'm happy to merge this now.

latchr:
Can we now delete the savelogic_improvement branch, or should there be further development happening there?

PR #179, opened 2017-04-02
remove fysom event parameter from on_activate and on_deactivate

tobiasgehring:
This patch removes the fysom event parameters (usually called "e") from on_activate and on_deactivate methods.

Reason
I couldn't find a single qudi module which makes actually use of this parameter. This patch thus simplifies this matter.

What if?
If the fysom event is needed at some point in some specific module, it is possible to override the default callback for onactivate and ondeactivate events

### DIFF
--- a/core/base.py
+++ b/core/base.py
@@ -69,8 +69,8 @@ class Base(QtCore.QObject, Fysom):
             callbacks = {}
 
         default_callbacks = {
-            'onactivate': self.on_activate,
-            'ondeactivate': self.on_deactivate
+            'onactivate': lambda e: self.on_activate(),
+            'ondeactivate': lambda e: self.on_deactivate()
             }
         default_callbacks.update(callbacks)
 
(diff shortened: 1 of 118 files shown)


drogenlied:
I disagree, they may not be currently used but have ben helpful when debugging module activation and deactivation on multiple occasions.

tobiasgehring:
if you want to debug, you can simply add one line to the constructor, as
well as the parameter. One line for debugging versus adding this
parameter to every module (including a five line documentation) and
explaining it to newbies.

tobiasgehring:
To be a bit more explicit, for debugging you need to change the
super().__init__() call in the constructor of the module to

super().__init__(self, callbacks={'onactivate': self.on_activate}, **kwargs)

and add the parameter to the event handler:

def on_activate(self, e):
   ...

You *can* actually add the parameter permanently, by defining
def on_activate(self, e=None):
which is already done sometimes. But you don't *have to*.

drogenlied:
Ok, that is sort of convincing.

Can you add the debugging explanation to `documentation/statemachines.md`?

Also, this warrants a notification to people using out-of-tree modules.

tobiasgehring:
I added the documentation and agree with the notification

PR #181, opened 2017-04-10
Extensions

tobiasgehring:
This patch introduces extensions to qudi. An extension is a python package with the regular folder structure with gui, logic and hardware. Each module defined in a qudi extension is added to qudi's gui/logic/hardware namespace. Here an example:

- qudi-extension
  - gui
  - logic
  - hardware
    - my_new_oscilloscope.py

After adding the extension to qudi the hardware module my_new_oscilloscope is available as
`import hardware.my_new_oscilloscope`

To make this possible implicit namespaces introduced in python 3.3 are used. This requires that `__init__.py` files are *not* used throughout the gui/logic/hardware packages - neither in qudi nor in an extension.

To add a qudi extension currently two different ways are possible:

1) in the configuration file
[global]
extensions = ["path to extension 1", "path to extension 2"]

2) use the PYTHONPATH environment variable

Please note:
- since the modules in qudi extensions and the regular modules shipped with qudi share the same namespace, name clashes may lead to unexpected behavior. Name clashes are solved by the search order defined in `sys.path` (i.e. the order defined in the extensions key in the configuration file, and the first match is included. That might be the biggest disadvantage. On the other hand it is simple to include modules from other extensions and it is always possible to define subpackages within an extension.
- It is possible to define only one of the three namespaces in an extension.
- Maybe we should run a check for `__init__.py` on an extension before including it to avoid troubles.
- Documentation about extensions is missing

Please share your thoughts.


tobiasgehring:
sorry for PEP8ifying core/manager.py in the same commit. Only real change is the inclusion of the global key "extensions" in the configuration file.

drogenlied:
Apart from the missing documentation and the fact that the maximum line length for this project is 100 characters (see [documentation](https://ulm-iqo.github.io/qudi-generated-docs/html-docs/programming-style.html)), I see no issues with the extensions concept.

tobiasgehring:
I wasn't aware of the 100 characters, sorry (I prefer 80 for practical
reasons on my machine). If you prefer I can copy the new extension code
to a new branch and make a new pull request. I'm working on the
documentation in the meantime.

PR #187, opened 2017-04-20
Signal signature

a-stark:
This branch contains first the fix for the signals, where None is not possible to use any more. 
Moreover some small modifications where done on AWG5002C, to make it work and the microwave hardware SMR20 was integrated in Qudi.
Two small bugs were fixed in in sequence generator, which appear on windows os.
Remove unit prefix from logic modules.
For more info see commit message.


a-stark:
Thanks for the comments.
Concerning the first point 'basic_extraction_methods.py line 29', I actually obtain an error in 
 'pulsed_measurement_logic.py line 768' since the return type is not a dict any more, but now an array. 
If I check the doc string of the method 'gated_conv_deriv' in 'basic_extraction_methods.py' then the return type should be of type 'dict' and not 'np.ndarray'. I think this was not desired, or am I wrong?

Concerning the second point. Be aware that we are using the scientific spinbox, that means that is a relative value. The smallest step size is actually 0.0001V. Can you please check whether you could reproduce that? Type in 0 for the amplitude, enter, then select again the cell and scroll with the mouse wheel up. That should move by the smallest step size (which should be 100micro=0.0001).

Neverhorst:
Ok, the stepsize of the spinboxes is indeed 100micro. But what is that parameter supposed to do then? This is not directly related to this pull request but shouldnt we be able to set a proper stepsize in scientific spinboxes? Kind of a lower bound that still makes sense (for example in the lower percentage bound of the confocal colorbar where one step is something like micro%). So that's fine in that case.

And yeah. At second thought there has been a merging conflict some time ago related to the pulse extraction that had to be resolved by hand... The ndarray return values are obvious residues of that problem. You are of course right that the extraction method should return a dictionary with the extracted laser pulses, the indices of the rising flanks and the indices of the falling flanks (or just one index in case of a gated extraction). Somehow this has been requested for singleshot measurements although I think it's currently only used in external standalone modules. 
So instead of simply not returning at this point, a proper dictionary with only zeros as laser_data should be returned early (see "return_dict" some lines further down).
The whole idea is that the user will not see "evaluated" data in the GUI if the extraction or analysis went wrong since this only leads to confusion.

a-stark:
Ok, the return will be now an array of zeros of extraction failed and return type corresponds now to docstring.
It is completely true that the 'view_stepsize' parameter has to be removed from the logic, since it is just gui related (that is a remain of the first version of pulsed). 
I do not want to make this pull request more chaotic as it is already, therefore I did not include this change here, I will address this in a separate commit/push.

Neverhorst:
Ok, thank you. I just saw that all other extraction methods also still have wrong return values (ndarray and not dict) when returning early during sanity checks. Let's merge the current status now and maybe you could change this in the other extraction methods as well later on? Should be quick and easy.

a-stark:
sounds good, I will also have a look at the other extraction methods.

PR #188, opened 2017-04-20
Check necessary imports

a-stark:
Check upon start of Qudi whether vital method are present to be able to start it and provide useful information on how to get missing packages.
This should help new people to start at least qudi if conda environment installation fails or was not performed correctly.


a-stark:
Vital packages are 'ruamel.yaml', 'rpyc', 'fysom'. I do not consider pyqtgraph as vital, rather optional. However, without pyqtgraph no MainGUI is displayed. How do you think about pyqtgraph?

Neverhorst:
If qudi starts without pyqtgraph and its logic modules are useable, then pyqtgraph is not "essential". However if I recall correctly the manager GUI won't start without pyqtgraph, right? If that is the case then I think this module is mandatory. That's at least my opinion.

a-stark:
You are right, the manager GUI will not start without pyqtgraph, therefore I am not sure how you could use qudi without the manager. 
In principle, if you change your config file such that on startup the kernellogic from the jupyterkernel is loaded instead of the manager GUI, then you could control qudi via the notebook. From the perspective of a new qudi user, this seems to be quite advanced.
Consequently, if I do not see the managerGUI with the default config, then I cannot use qudi and from this pyqtgraph seems to be vital. Currently I do not see another way to circumvent the problem.

drogenlied:
I think Tobias spent quite a lot of time to make Qudi run without pyqtgraph and refactored the core parts so that they run on qtpy only and not on pyqtgraph.

There are provisions in gui/manager/managergui.py (see `_has_pyqtgraph`) that deal with not having pyqtgraph.
The fact that this has not been tested sufficiently and breaks is no reason to require pyqtgraph.

The main issue with pyqtgraph right now is that it is effectively unmaintained but there is no viable replacement.

PR #189, opened 2017-04-21
Qo fastcounter rev2

Neverhorst:
Overhaul of the qo-fastcounter. Bitfiles have been changed in thirdparty repo and so did the hardware module in this branch. Tested carefully for both FPGA types (XEM6310-LX45 and XEM6310-LX150) with signal generators only and on a real setup as well. Now no strange timetrace features should show up anymore and the LX45 FPGA is useable.


drogenlied:
Can you convert these huge power-of-two numbers to hexadecimal? That makes it way easier to see which bit is set.

Neverhorst:
Done. You are right, this looks much better. All numbers are 8-digit hex numbers now.

PR #191, opened 2017-04-24
Improve Appveyor build

drogenlied:
* use tools/conda-env-win8-qt5.yml to install known-good python environment
* no more python 3.4/3.5 Qt4/Qt5 tests. 3.6 is official now.
* produces log and zipped test notebooks as artifacts
* get rid of astropy ci-helpers
* simplify .appveyor.yml


drogenlied:
I think we already dropped Qt4 (or pyqt dropped it for us because the latest version produces errors).

PR #194, opened 2017-04-26
Ungated threshold extraction fix

Neverhorst:
Fixed the currently not working "ungated_threshold" extraction method. This was a not-reviewed relic from pi3d and was performing so poorly that the analysis loop caused qudi to hang (at least the corresponding logic thread). Rewrote this method completely using numpy.
Currently the "Tolerance (bins)" parameter is not used. Formerly it was even used in a wrong way. This feature can be implemented later by adjusting solely the extraction method. The input widget remains in the GUI but is disabled until someone implements this feature.


Neverhorst:
Oops, sorry. Done.

PR #211, opened 2017-06-06
Microwave interface overhaul

Neverhorst:
Quite a lot of changes for the whole "ODMR toolchain".

The hook for those changes was the fact that half of the microwave devices in qudi was not properly implemented to avoid confusion between CW, List and Sweep mode (on a device level) when changing between them. The reason for that behaviour was a microwave interface that had no clear direction in that sense. This probably is a result from a more or less direct port from pre-qudi times.

I tried to fix that a bit by distinguishing strictly between cw, list and sweep mode. The hardware module should behave like each one of those modes has it's own parameter set (freq and power for cw mode, freq_list and power for list mode, start/stop/step and power for sweep mode) and can be activated at any time. The method set is in principle:

- off()
- set_cw(frequency, power)
- set_list(frequency_list, power)
- set_sweep(start, stop, step, power)
- cw_on()
- list_on()
- sweep_on()

So you can for example set parameters with set_cw() then set other parameters with set_list() and afterwards call cw_on() using the parameters previously set.

This changeset made it neccessary to alter the microwave_interface.py file. And this also results in changes in all microwave hardware modules. I tried to test all these changes on real hardware but was limited to Anritsu, Anritsu70GHz, SMIQ, Gigatronics and of course the dummy module.

**SMIQ and Gigatronics still need to be tested with a triggered frequency sweep (list and sweep mode).**

In the process I also adjusted/enhanced ODMR logic and GUI modules a bit. The most important changes in that respect are:

- separate power SpinBoxes for cw mode and list/sweep mode
- removal of the ComboBox for toggling of cw mode (new action in toolbar instead)
- odmr raw data is now always saved when the user is saving it. The truncated mini-matrix had no relevance and is only used to plot a picture of the measurement.

All in all I hope this should give a solid base to refine these parts of qudi even more without touching the interface again.

**Please briefly test your microwave output in list/sweep and cw mode with an oscilloscope or a spectrum analyzer. Everything should be fine but you never know.**

**Also Agilent, SMR20 and srssg devices will not work (anymore?) since I don't have them here. Please adapt them to the new interface if you have them.** From the looks of it some of them should not have worked even before this update.


drogenlied:
Let's keep this here until Friday and see if anyone manages to complain.

Neverhorst:
Well, since there are apparently no complaints... let's merge.
Please create tickets for future problems.

PR #214, opened 2017-06-13
Fysom update

tobiasgehring:
This patch makes use of the latest fysom version 2.1.4 and includes several improvements
- removal of FysomAdapter and related __getattr_ implementations
- GUI modules can now directly inherit from QMainWindow (as an example I converted the manager GUI)
- removal of _wrap_activation and _wrap_deactivation
- added version checks for required python packages. So far only fysom got a required version number.

Alexander Stark has tested this patch already, but please test it further as it concerns the very base of qudi.


a-stark:
Thanks @tobiasgehring for the pull. I think it would be good if you could test that on at least two different systems (e.g. a debian and ubuntu maschine or windows 10). I have tested it on windows 7 OS in the lab, so far all the measurements were running normally.
As @tobiasgehring has mentioned, it is actually a small change but has a large impact how qudi is handling multiple inheritance. This commit would then also constitute ultimately the end for qt4 support, but it has been dropped anyway, so just that you know.
There seem to be some problems in Travis CI and AppVeyor, in the fit testing part, but I think that has to be solved in a different pull request.

drogenlied:
@a-stark CI fails because the fit functions fail horribly with the given test cases. The only difference before failure and after failure is the addition of tests.

@tobiasgehring, this is really nice and it works on Debian when Fysom is installed from pip, but it comes at an unfortunate time as I have a branch that uses exactly _wrap_activation and _wrap_deactivation to enable a declarative style for configuration values, status variables and connectors that does some behind the scenes work and replaces the declared class variable during activation and saves status vars after deactivation.

I would merge this anyway, as I am sure I can find a way to replicate the functionality with the new system, but some advice would be appreciated.

Let me check this out on a few lab computers before approving it.

tobiasgehring:
@drogenlied, can you post some code you execute in the wrappers or
better give me access to your branch so that I can have a look?

drogenlied:
The branch is public: https://github.com/Ulm-IQO/qudi/tree/explicit-config-options
The activation and deactivation wrappers look like this:

```
    @QtCore.Slot(result=bool)
    def _wrap_activation(self):
        """ Restore vars and catch exceptions during activation. """
        # add status vars
        for vname, var in self._stat_vars.items():
            if var.name in self._statusVariables and var.type_check(self._statusVariables[var.name]):
                setattr(self, var.var_name, self._statusVariables[var.name])

        self.log.debug('Activation in thread {0}'.format(QtCore.QThread.currentThreadId()))
        try:
            self.activate()
        except:
            self.log.exception('Error during activation')
            return False
        return True

    @QtCore.Slot(result=bool)
    def _wrap_deactivation(self):
        """ Save vars and catch exceptions during deactivation. """
        self.log.debug('Deactivation in thread {0}'.format(QtCore.QThread.currentThreadId()))
        try:
            self.deactivate()
        except:
            self.log.exception('Error during deactivation:')
            return False

        # save status vars
        for vname, var in self._stat_vars.items():
            if hasattr(self, var.var_name):
                self._statusVariables[var.name] = getattr(self, var.var_name)

        return True
```

tobiasgehring:
I suggest to replace the default handler for on_activate and
on_deactivate in the constructor,

 default_callbacks = {
            'onactivate': lambda e: self.on_activate(),
            'ondeactivate': lambda e: self.on_deactivate()
            }

by your own function which then calls self.on_{de}activate().

Are you sure your setattr does never fails or overwrites something?
Sounds dangerous.

drogenlied:
My setattr always overwrites something, namely a class variable of type StatusVar  that has been defined in the module beforehand and is detailing what value it should be overwritten by.
I know it is of this type because the type is checked beforehand when creating self._stat_vars.

Overwriting this should never fail, except in cases where somebody messes around with metaclassing.

Should overwriting fail, all bets are off anyway as there is now an object of the type StatusVar in the place where the content of said status variable should have been.

tobiasgehring:
I was more wondering about somebody naming a status variable "__init__",
but I have to admit that I didn't carefully inspect your code yet.

drogenlied:
For that scenario, somebody would have to explicitly write
```
class XYZ(Base):
    __init__ = StatusVar(...)
```
which is obviously a bad idea.

tobiasgehring:
ok, perfect. :)

drogenlied:
There are some consequences that I missed due to focussing on my status variables issue:

 * Running activation in the main thread breaks modules that start a QTimer during activation, and there are a few of those. Also, having objects lying around that belong to a different thread/event loop leads to strange behaviour.
 * `nuclear_operaions` GUI has a constructor that is very oldschool and just breaks (@a-stark !!):
```
Error while loading gui module: nuclearops
  File "/home/jan/qo/qudi/core/manager.py", line 737, in loadConfigureModule
    self.configureModule(modObj, base, class_name, key, defined_module)
  File "/home/jan/qo/qudi/core/manager.py", line 571, in configureModule
    instance = modclass(manager=self, name=instanceName, config=configuration)
  File "/home/jan/qo/qudi/gui/nuclear_operations/nuclear_operations.py", line 59, in __init__
    c_dict)
TypeError: QObject(parent: QObject = None): too many arguments
```

tobiasgehring:
thanks for pointing out the QTimer, this slipped @a-stark 's and my
mind. I was thinking that as long as the parent is the thread it
wouldn't matter in which thread the object was created.

I have to think about how to fix this. The problem is that invokeMethod
as it was used before does not work with slots defined in mixins. It
simply doesn't recognize the method. Any ideas?

tobiasgehring:
what about activation first, moveToThread later? That should move the QTimer (if it is a child of the module object) to the new thread as well. Any comments?

drogenlied:
I'm not sure how well the Qt object ownership works with pyqt.

The only obvious place where invokeMethod could be used are the functions that I added to save and restore status variables.

The distinction between logic modules that need to move and the others can be done by checking for `getModuleThread()`.

Slots in mixins should be possible according to the PyQt docs (http://pyqt.sourceforge.net/Docs/PyQt5/pyqt4_differences.html):
> Unlike PyQt4, PyQt5 supports the definition of properties, signals and slots in classes not sub-classed from QObject (i.e. in mixins).

drogenlied:
Also, did you try explicitly decorating the function as a Qtcore.Slot()?

drogenlied:
See https://github.com/Ulm-IQO/qudi/commit/f7599ef57e4dbc75b91eefabaa433ab3bd4ccecc for a fix.

The fix is not totally straightforward as you need to apply a `Slot()` decorator and take a detour through `trigger(event, ...) ` as `activate()` is generated at runtime and `invokeMethod()` does not find it.

tobiasgehring:
nice work, thanks! I was wondering if it is possible to make sure it's
always run in the right thread if module.activate() or
module.deactivate() is called. I guess it could be done with a check for
the current thread in an overridden trigger() method that calls
invokeMethod on itself? But may that's a bit overkill?

Just wondering: what happens if someone wants to run a hardware module
in a thread? Is that possible and has somebody thought about it before?

drogenlied:
At first, I was thtinking about adding the invokeMethod to my `statusVariable` restoration functions, but then that creates a sort of logical dependency where an instance of the parent class `Base`/`BaseMixin` needs to check if it is also an instance of a child class, `GenericLogic`.

I would also like to add you to the Github project as a collaborator, as I am preparing a pull request for my config options/status variables branch and would like to hear your opinion on the design.

PR #215, opened 2017-06-20
Static and explicit connectors, configuration options and status variables

drogenlied:
This branch introduces a way to explicitly declare connectors, configuration options and status variables.

* This is a prerequisite to build a graphical configuration tool.
* It also makes status variable saving and restoring happen automatically for declared variables in the trivial case.
* It makes loading configuration options easier in the trivial case and exposes them statically so they can be used in external tools (see first point) and adds error messages or hstops module loading when required values are not present.

This is achieved by means of a new metaclass, `ModuleMeta`, that bundles all `Connector`, `ConfigOption` and `StatusVar` objects into a dictionary for later use in `BaseMixin`, etc.

Declaring connectors with `_in` and `_out` does not work anymore, but declaration via `_connectors` still works, even though it is not preferred.

Most or all of the cases where Connector, ConfigOption and StatusVar usage is trivial have been converted, the non-trivial use cases that are not possible yet continue to work the old way while a replacement is actively worked on.

Furthermore, some things that were not really planned for ended up in this branch:
 * Fixes errors introduced in Fysom update merge
 * Converts most uses of pyqtgraph ImageItems to row-major data format


drogenlied:
Tests fail because nobody has fixed the fit functions yet.

Neverhorst:
Sorry, I can only test it next week...

tobiasgehring:
I like the concept and the implementation very much, thank you for bringing this forward. I have a couple of remarks which hopefully will help to further improve it:

**connectors**
```
class Module(...):
  myconnector = Connector(interface_name='SaveLogic')
  def on_activate():
    self._my_connector_object = self.get_connector('myconnector')
```
- it would be nicer if the `get_connector()` call wasn't necessary anymore. Ideally `self.myconnector` should point to the connected object (or None if not connected). Also ok would be `self.myconnector().do_something()`
- `interface_name` is so far not really used and it's only purpose is to have a configuration tool. I suggest to point to the actual interface definition, i.e. `myconnector = Connector(interface=SaveLogic)` and use it to check it during the establishment of connections.

**StatusVars**
- in the counter_logic I found that `counting_mode` is handled differently to other status variables. The reason is that it needs some kind of conversion from the text file value to an object. I would prefer if status variables are handled in one way and not in different ways depending on usage. I thus suggest to add two parameters to the StatusVar class which can be used to do the conversion (having the module and the value as parameter):

```
myvariable = StatusVar('myvariable', 300, getter, setter)
```
where getter is a callable that converts from the config file value to the modules interna and setter the other way around.
- I find it a bit cumbersome that internally a _statusVariables OrderedDict is used which is synced in activation and deactivation event handlers. It would be nicer if it uses a new save and load implementation not going this detour.
- it's not completely clear to me how type_check() is used.

**ConfigOptions**
I didn't look closely at config options yet.

small cosmetic issues:
- please have two blank lines before class definitions to be compliant to pep8: https://www.python.org/dev/peps/pep-0008/#blank-lines
- StatusVar.type_check not documented

tobiasgehring:
can you explain and comment on the crashes you saw with ManagerGUI being a subclass from QMainWindow?

drogenlied:
1. Getting rid of `get_connector` is the end goal, but I feel that I got into too much trouble with this branch already to simply do that replacement right now, even though it should be easy.
 2. `interface_name` same point, the use of it is basically to keep things the same as before the change
 3.  the old `_statusVaraibles` is still around because removing it requires actual work on each module where there is some sort of special case handling. The setter/getter pattern only works with statically declared getters or setters, which do not cover all of the cases present in currently existing modules. If they are dependent on things that a module will only know at runtime, we can forget ever having a non-text configuration tool.
4. `type_check` is a weaker form of a setter, but as you proposed a setter/getter pattern, I would just rework it into that.
5. The crashes with the converted ManagerGui happen on a certain computer in our lab on 64bit Windows 8. The crashes are randomly either WindowsExceptions or memory access violations. The WindowsException case is not tied to any thread, the segfault happens in the main (GUI) thread at `super().show()`, line 213 in managergui.py. Preventing the `show()` function fom being called always produces the WindowsException case.

drogenlied:
Hmm, about these blank lines, I thought I had those everywhere but my vim setup is obviously not good enough for me to notice if they are missing, the symbol is the same as for a multitude of other PEP8 nitpicks for a certain class.

tobiasgehring:
1. what about offering get_connector additionally to keep the old behavior?
2. you could implement both and add checks if it is a class and not a string. And then at some point deprecate the string. Not adding it now makes it completely unclear when a good time would be.
3. can you give an example? I'm not sure I understand these use cases. A setter/getter can still react on run time information as it will get access to the module object. I'm not sure what a configuration tool has to do with status variables, doesn't that only effect ConfigOptions (or are you referring to a similar problem there)?
I'm also wondering about making a _statusVariable OrderedDict subclass that directly writes changes to the StatusVars to achieve backwards compatibility that can easily be removed later. (maybe too much work)
5. damn. 
6. We should more clearly mark deprecated parts in the code so that it doesn't become a huge mess.

drogenlied:
I will see what I can do for 1. and 2. today.

3. does not make sense because I managed to get `StatusVar` and `ConfigOption` mixed up. You are right that only `ConfigOption` needs to be static and it needs a way to a) show what values it acceps and b) check these values.

Getting rid of _statusVariables right now is effort for little gain, porting the remaining modules to use setters/getters and then just removing it looks more promising.

drogenlied:
Do you have a better idea than passing a function name as a string for getters/setters?
See 7d6d00d for an example.

tobiasgehring:
a possibility would be to define the getter / setter above the status variable and use its method name. However mixing up the declarations and the setter / getter implementations is not so nice as one easily looses the overview of available status variables.

Another possibility is to not define getter and setter in the StatusVar() call at all and use decorators to define them instead. I haven't completely thought it through yet, but I mean something like:

```
class ODMRLogic(GenericLogic):
    fc = StatusVar('fits', None)

    @fc.setter
    def sv_set_fits(self, value):
        pass

    @fc.getter
    def sv_get_fits(self):
        pass
```

If it works, the obvious disadvantage is that you cannot see from the StatusVar declaration that the status variable has getter and setter functions. Not sure that's necessary though.

tobiasgehring:
did you implement the getter / setter functionality in some module to try it out?

drogenlied:
Yes, check `logic/odmr_logic.py`. I'm not sure if that is how you envisioned it, but it works.

tobiasgehring:
Thanks. Yes, it's about what I imagined.

I'm wondering whether we should implement something like get() next to __call__() for the connectors. Might be easier to grasp by some people. (I though would keep the __call__ implementation)

```
self._taskrunner = self.taskrunner()
self._taskrunner = self.taskrunner.get()
```

tobiasgehring:
would it make sense to add a setter for the ConfigOption as well?

drogenlied:
The call style for attributes is typical Qt behaviour, I would not worry too much about it.

And yes, it would make sense to add a setter to ConfigOption.

I am in Singapore for about two weeks and do not have the usual amount of time.

tobiasgehring:
I have started to write a fix for #223 and realized that a different implementation of the setter / getter functionality for the status variables might be a good idea. Right now it is implemented in the way that on deactivation of a module possible getter functions are executed with the intention of: "give me all status variables you have and want me to save!" Rather than that I would prefer a more status variable centric approach. Please have a look at the branch [status-variables-immediate-saving](https://github.com/Ulm-IQO/qudi/tree/status-variables-immediate-saving). Here a status variable is explicitly assigned and the getter / setter functionality does nothing else than converting data types. So in the odmr_logic it is converting a FitContainer to an ordereddict and vice versa. However, the actual assignment of FitContainer -> status variable is done in on_deactivate and status variable -> FitContainer in on_activate.

drogenlied:
That discussion should not be in this pull request.
Please post to the mailing list or create a new pull request.

tobiasgehring:
I disagree. My intention was to suggest to change the implementation of this pull-request.

drogenlied:
I disagree with your disagreement.

This pull request should not contain any change in behaviour, so even though things work differently under the hood, the actual behaviour, even of things that are now considered legacy, does not change.
The old ways of using `_statustVariables` or declaring connectors with `_connectors` continue working.
The threading behaviour is unchanged compared to before this pull request.

Apart from these intentional reasons, there is something very wrong with this approach if I understand it correctly:

[Disk access](http://i.imgur.com/k0t1e.png) is [extremely slow.](https://gist.github.com/jboner/2841832)
Latencies for a single read or write are at least 10 milliseconds.
Concurrent access can push file operations into the seconds range.
Sequential reads can yield 100MByte/s in the best case but disk access from other programs can drop this to 100kbyte/s or even lower.
Caches hide a lot of this behaviour, but simple things like the MS Windows update service can lead to constantly full caches. Under these circumstances, opening a file browser window can take in the order of minutes.

Writing all status variables to disk when one of them is changed seems like the perfect storm for a completely laggy or hanging application.
Let me take the Confocal component as an example:

The Confocal cursor position is stored as a status variable.
The image history of confocal scans, including all relevant settings, is also stored as a status variable.
Through a nice trick, the image data is written to separate, compressed files using numpy functions.
This was necessary because a 200x200 scan is not really small and a 1000x1000 scan was pushing past the limits of what the ruamel.yaml parser can handle as inline data.
This means that, with your implementation, simply moving the cursor in Confocal can write tens to hundreds megabytes to disk, distributed over 10 files or whatever value the scan history length is set to.

Depending on which thread the module runs in, this can just make the logic module unresponsive and or hang the whole application. Right now, savelogic does something like this, but that tends to be triggered by an explicit and rather rare click of a save button and not on literally every single action.
Even the in the case of savelogic, this behaviour is highly annoying.

You can discuss the disk speed issue in detail with @Neverhorst when he is back from the US in August, he has some stron opinions on this topic.

In the meantime, I would rather not have this change in the code.

tobiasgehring:
@drogenlied: ok, now I know why we disagree :) It's just a misunderstanding. I'm very very sorry for that. The discussion you brought up does in fact not belong here. I will shortly open a pull request where we can discuss these issues.
Let me please try again and please forget everything I wrote above.

Status variable declaration and usage as implemented:
```
class MyNewModule(Base):
    my_status_variable = StatusVar('my_variable', 50)

    def read_example(self):
        # status variable read
        return 30*self.my_status_variable

    def write_example(self, x):
        # status variable write
        self.my_status_variable = 30*x + 3
```

So far so good. Now we have setter and getter functions:

```
class MyNewModule(Base):
    my_status_variable = StatusVar('my_variable', None)

    @my_status_variable.setter
    def setter(self, val):
        # here we basically read the value of my_status_variable and do something with it;
        # the value is contained in the parameter val
        self.fc = construct_some_object(val)

    @my_status_variable.getter
    def getter(self):
        # take what is in self.fc and return in a format that can be stored in the config file
        return serialize_in_some_way(self.fc)
```
Here, `my_status_variable` is in fact never accessed directly. The setter and getter functions are only called in the wrappers around `on_activate` and `on_deactivate`.

My suggestion is a usage like follows:

```
class MyNewModule(Base):
    my_status_variable = StatusVar('my_variable', 50)

    @my_status_variable.getter
    def getter(self, val):
        # here we convert the value of my_status_variable stored in the status file and convert it to our own type
        return construct_some_object(val)

    @my_status_variable.setter
    def setter(self, val):
        # here we take the object and convert it into a format that can be stored in the config file
        if not isinstance(val, MyConstructedClass):
            raise Exception('Invalid type for status variable.')
        return serialize_in_some_way(val)

    def on_activate(self):
        # convert status variable to object
        self.fc = self.my_status_variable

    def on_deactivate(self):
        # convert object to status variable
        self.my_status_variable = self.fc
```

I think this is a much more consistent and cleaner way of using getter and setter functions. They are merely a converter when reading and writing status variables. Of course one would like to cache the object and not use the status variable directly all time, like in the example above, where this is done in `on_activate` and `on_deactivate`.

I hope my idea is much clearer now. I apologize again for having been confusing.

drogenlied:
That is much more reasonable, but I fear that while it is nicer from a design point of view, it will just make implementing the unusual cases more difficult.

The fit container example is one of the most straightforward one-to-one mapping examples, where one status variable corresponds to one module variable and there are very few checks or dependencies.

Requiring the proposed "purely functional" behaviour will require redesign, which is sort of unpopular work as it tensds to break things in unforeseen ways.

tobiasgehring:
I can understand your concern, however I believe that it can be solved in another way. Everything you execute in the current implementation of the getter and setter can be implemented in `on_activate` and `on_deactivate` as well.

Remark: if you follow an implementation like in the `status-variables-immediate-saving` branch the internal `_statusVariables` dictionary is always up-to-date.

drogenlied:
I would not be so sure about that. Have a look at `logic/confocal_logic.py` and carefully evaluate how lines 313 to 340 and the corresponding deactivation would translate to this concept.

tobiasgehring:
well, the problem here is that a previous design decision was to make the length of the history adjustable and not to put all entries into a list but into the root of the config file. In my opinion this is not compatible with the StatusVar concept at all, no matter which way the getter and setter methods are implemented. The concept you designed requires that all root entries are represented by StatusVar objects.

A workaround always will use the underlying `_statusVariables` dictionary. For example an implementation could use setter and getter functions on a virtual `history` status variable and access the individual entries using the `_statusVariables` dict. Below you find a completely untested version of the "new-style" getter and setter functions:

```
    virtual_history = StatusVar(None)

    # recover (`self.history = self.virtual_history`) and set (`self.virtual_history = self.history`) in on_activate and on_deactivate as in the example I posted above. Don't forget to call restore() in on_activate on an confocal history entry.

    @virtual_history.getter
    def virtual_history_get(self, value):
        history = []
        for i in reversed(range(1, self.max_history_length)):
            try:
                new_history_item = ConfocalHistoryEntry(self)
                new_history_item.deserialize(
                    self._statusVariables['history_{0}'.format(i)])
                history.append(new_history_item)
            except KeyError:
                pass
            except OldConfigFileError:
                self.log.warning(
                    'Old style config file detected. History {0} ignored.'.format(i))
            except:
                self.log.warning(
                        'Restoring history {0} failed.'.format(i))
        try:
            new_state = ConfocalHistoryEntry(self)
            new_state.deserialize(self._statusVariables['history_0'])
        except:
            new_state = ConfocalHistoryEntry(self)
        finally:
            history.append(new_state)
        return history

    @virtual_history.setter
    def virtual_history_set(self, value):
        if not isinstance(value, list):
            raise Exception('Wrong type for status variable virtual_history.')
        hist_index = 0
        for state in reversed(value):
            self._statusVariables['history_{0}'.format(hist_index)] = state.serialize()
            hist_index += 1
        return None
```

You could also simply leave it as is as you decided to do as well in this pull request. The rotating of the history is here actually included in the getter. That might not be a good idea and should rather be implemented elsewhere (e.g. in on_activate).

drogenlied:
Yes, that is a way to hack around the limitations of the "purely functional" getter/setter design.
This example could be even better if `history` was assigned directly, without the boilerplate code in `on_activate` or` on deactivate`.
Getting rid of repetititve statements in activation/deactivation was one of the better side effects of statically declaring things as it shrank the necessary code lines by at least a factor of 3.

While I now thing that this way can work nicely if done right, the end goal is still a point of contention.

Maybe I should generalize my previous statement about disk access to all sorts of heavy actions through getters/setters or `@property`.

We do not use `@property` for accessing the internal state of measurement devices, even though this would be perfectly possible and I think at least one of the instrument libraries in your review does exactly that. Why is that so?
Because people do not expect variable access to take any time at all while instrument access is as bad or worse than disk access.

tobiasgehring:
what about renaming `setter` and `getter` into `constructor` and `representer`? Terminology is borrowed from the yaml implementations.
In contrast to the `setter` function the `constructor` actually returns the constructed object. 
In `__load_status_vars_activate` we can then do
```
            if var.constructor is None:
                setattr(self, var.var_name, svar)
            else:
                setattr(self, var.var_name, var.constructor(self, svar))
```
and get access to the constructed object via the status variable. (I think this will also work in confocal_logic, though you still need to hack when using the current representation)

We can then think about a nicer implementation using the new API of ruamel.yaml once it's finished.

drogenlied:
Sure, we can do this, although I doubt anybody except you knows the ruamel.yaml terminology / API or can derive any additional meaning from the names.

PR #218, opened 2017-06-22
Templates that provide a guideline for bug reports and pull requests

drogenlied:
These templates provide a structure for submitting helpful bug reports and pull requests.


a-stark:
Very nice one.

PR #231, opened 2017-07-19
Fix jupyter notebook concurrency

drogenlied:
Fixes a privately reported bug where only one cell from a jupyter notebook can be executed at the same time, from different notebooks running on different kernels.

## Description
Give each QZMQKernel their own thread to execute in, instead of executing all of them in the jupyteḱernel module thread.

## Motivation and Context
Checking partial results of a running experiment requires concurrent execution of multiple notebooks:
One to run the experiment in and one to plot or calculate the results.
This was not possible, as only one notebook cell could run at the same time per Qudi instance.

Bug was reported privately

## How Has This Been Tested?
Run Qudi with qudikernel set up as per installation guidelines.
Open two notebooks, one with a cell containing a long sleep, one with a short print or calculation.
Run the sleeper cell, then run the short calculation. The calculation should return as quickly as it completes, while the sleeper cell should take the specified amount of time.

## Types of changes

## Checklist:

- [x] I have checked that the change does not contain obvious errors (syntax, indentaion, mutable default values).


a-stark:
@drogenlied, it seems that you have checked out by accident from the 'explicit-config-options' branch, rather from the master and applied to that your fix. I assume that the 'explicit-config-options' branch is not a prerequisite for the fix.
Could you rather check out from master and commit in this branch the changes (basically your last commit) and force push it to this branch, to exclude the changes from 'explicit-config-options'.
This will make the pull request much cleaner.
Apart from that, the fix looks reasonable. I could test it if you apply the changes mentioned above.

drogenlied:
No, lets get the config options in first. I am getting quite impatient with that thing.

PR #232, opened 2017-07-23
[WIP] saving of status variables

tobiasgehring:
I open this pull request with a work-in-progress tag to enable design discussions.

## Description

Status variables are so far only saved when a module is deactivated. If qudi or the operating system crashes before, all status variable changes are lost.

To prevent this situation this pull requests extends the `StatusVar` class by a new parameter `save` which controls when a status variable is saved to file. The default value is `deactivation` which is the previous behavior, so no behavioral changes on existing code.

The two new options are:
- `immediately`: saving of status variables is issued directly after the status variable is changed
- `delayed`: status variables are saved after a certain amount of time (currently 500ms) after the status variable is changed. If it is changed in the mean-time, the timer is restarted.

If one of the two new options are used, all status variables of a certain module are saved to file. This is important to note as this behavior has to be considered if huge numpy arrays are stored in status variables. Currently I don't see a way around this unless status variables are split over several files.

Other changes:
- status variables are implemented as properties, values are stored in the value attribute of the status variable
- I moved saving and loading of status variables from the manager to the module level. In principle a module could override the implementation if something else is needed.

## How Has This Been Tested?
- changed managergui to support saving the only status variable `delayed`.
- it definitely needs more testing.

## Types of changes

## Checklist:

### DIFF
--- a/core/manager.py
+++ b/core/manager.py
@@ -568,7 +568,7 @@ class Manager(QtCore.QObject):
                 instanceName, baseName, className))
 
         # Create object from class
-        instance = modclass(manager=self, name=instanceName, config=configuration)
+        instance = modclass(manager=self, base=baseName, name=instanceName, config=configuration)
 
         with self.lock:
             self.tree['loaded'][baseName][instanceName] = instance
@@ -912,7 +912,6 @@ class Manager(QtCore.QObject):
             logger.error('{0} module {1} not deactivated'.format(base, name))
             return
         try:
-            module.setStatusVariables(self.loadStatusVariables(base, name))
             # start main loop for qt objects
             if base == 'logic':
                 modthread = self.tm.newThread('mod-{0}-{1}'.format(base, name))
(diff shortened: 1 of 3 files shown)


Neverhorst:
Just wanted to ask what the status of this pull request is right now. Is it still WIP or is it ready for review?

As a remark one could also save in fixed intervals of say 5-10 min if a variable has changed. 
A potential crash within this time is not a big deal.

a-stark:
What do you think @tobiasgehring, do you consider to work further on this pull request? Otherwise we could close it, since there were no big changes since the last year.

tobiasgehring:
yes, it is still work in progress and I would like to finish it, but right now I don't know when. If you prefer we can also close the pull request but keep the branch until it is a bit further developed.

a-stark:
I would suggest closing this pull request until you have a working solution which can be checked. Would this be ok? But the branch will be kept for sure!

Neverhorst:
Closed until the branch has been further developed.

PR #252, opened 2017-08-27
Rpyc / Remote server optional

tobiasgehring:
So far rpyc was a core dependency and an rpyc server was always started even if the functionality wasn't used at all. This patch makes both optional. 

## Description
Using a remote module is a nice feature in qudi, however, not quite everybody is using it. Up to now rpyc had to be installed and a server was started even if the functionality wasn't used at all. This patch disables all remote module functionality if rpyc is not installed. If it is installed qudi can be used as client or as server or both, depending on configuration.

The server is now disabled if no corresponding configuration is available. This can in principle break existing configurations (though I hope that people using the functionality use certificates!).
New configuration options:
```
[global]
module_server:
  address: ''
  port: 12345
  certfile: 'filename.cert'
  keyfile: 'filename.key'
```
The old configuration options still work and are labelled as deprecated. The reason for the change is that I prefer to have them grouped together.

Note: This patch does not intend to fix any issues within the remote module implementation.

## How Has This Been Tested?
I tried the different configurations (server, client, both) including an uninstalled rpyc.

Would be great if someone actually using the functionality could check that it is still working as expected.

## Types of changes

## Checklist:

No documentation describing the remote module functionality exists yet.

### DIFF
--- a/config/example/default.cfg
+++ b/config/example/default.cfg
@@ -9,8 +9,9 @@ global:
     # list of modules to load when starting
     startup: ['man', 'tray', 'tasklogic']
 
-    serveraddress: localhost
-    serverport: 12345
+    module_server:
+        address: 'localhost'
+        port: 12345
 
     ## For controlling the appearance of the GUI:
     stylesheet: 'qdark.qss'
--- a/core/manager.py
+++ b/core/manager.py
@@ -35,13 +35,16 @@ import importlib
 from qtpy import QtCore
 from . import config
 
-from .util import ptime
 from .util.mutex import Mutex   # Mutex provides access serialization between threads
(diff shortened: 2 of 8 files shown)


drogenlied:
Apart from the missing documentation and the
`Remote module sharing does not work as server startup failed earlier, check your log.`
error that needs to be rephrased this is ok.

tobiasgehring:
- added documentation
- changed error message

tobiasgehring:
will you merge it? I'm not authorized to do this.

drogenlied:
Strange, the permissions in the interface are the same for you and the qudi-committers group.

PR #254, opened 2017-09-08
Odmr counter microwave interfuse

a-stark:
## Description

- Implementation of an odmrcounter-microwave interfuse.
- This enables the odmr counting by using software trigger, to jump to the next entry in either list or sweep mode.
- The counting is still based on hardware.
- Decrease in this configuration the amount of required counters in the hardware from 4 to 2.
- Introduce a new method (trigger) in the odmr_counter_interface. This method will be only used in the interfuse case.
- Implement this new method for all existing microwave sources to ensure compatibility
- Add to the default config an example how it will be connected.

Note, this branch has been rebase onto master.

## Motivation and Context

There are still existing devices, which possess no hardware input trigger to switch to next list entries (and I need to use unfortunately such an archaic device). The most necessary part: the counting is still hardware based but the trigger event originates now from a trigger event by the hardware module.

This principle enables not only to integrate microwave sources without hardware trigger but also reduce the amount of required counter channels on the NI card. As a result, just 2 counters are used, hence older NI cards (e.g. NI 6229) can now easily be used for ODMR counting.

The new method in the interface is implemented in all hardware files to prevent an exception upon start. In the normal ODMR mode, this method will not be used anyway, so no problems should appear here for existing use cases.
However, the trigger method was not tested on the hardware devices. If this counting scheme is needed, then the trigger method has to be checked for functionality (I included at least the command I found in the specs for each device)

The default config contains now an example how to use this interfuse. Since another dummy (slow_counter_dummy) is essentially used to create the counts, no odmr dip simulation could be implemented.

## How Has This Been Tested?

I cross checked the resulting ODMR traces with the SMR and the SRS (only interfuse config possible), and it works like a charm. The modular design shows its benefits here! The default config will have also a functioning dummy (but without ODMR dip simulation, see above why).

The only drawback of that scheme is obviously, that the communication with the PC limits the speed (but this is just in the ms range). For a configuration of clock frequencies of 100Hz, 200 frequency points (software trigger) I got the same results as with the SMR20 with 200Hz counting (also 200 frequency point with hardware trigger).

So keep in mind, you will not loose time if the clock frequencies is reduced say to 50Hz! You will solely count longer per point which increases your signal and reduces the triggers per time interval.

## Types of changes

## Checklist:

### DIFF
--- a/config/example/default.cfg
+++ b/config/example/default.cfg
@@ -9,8 +9,9 @@ global:
     # list of modules to load when starting
     startup: ['man', 'tray', 'tasklogic']
 
-    serveraddress: localhost
-    serverport: 12345
+    module_server:
+        address: 'localhost'
+        port: 12345
 
     ## For controlling the appearance of the GUI:
     stylesheet: 'qdark.qss'
--- a/config/example/default.cfg
+++ b/config/example/default.cfg
@@ -21,11 +21,10 @@ hardware:
     simpledatadummy:
         module.Class: 'simple_data_dummy.SimpleDummy'
 
-    mykrowave:
+    microwave_dummy:
         module.Class: 'microwave.mw_source_dummy.MicrowaveDummy'
         gpib_address: 'dummy'
         gpib_timeout: 20
-        mock: True
 
     mydummyscanner:
         module.Class: 'confocal_scanner_dummy.ConfocalScannerDummy'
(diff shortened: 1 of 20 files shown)


drogenlied:
You are using funtions from `self._odmr_counter` that are not part of the `ODMRCounterInterface`.
You need an additional connector for a `SlowCounter` because you are using its methods.

a-stark:
Oh boy, you are right. Actually, the connecter should also be renamed (both, the class and the instance attribute, odmrcounter and _odmr_counter, respectively). Thanks!

a-stark:
I think now it should be clearer.

PR #256, opened 2017-09-09
Pulsed save and display enhancement

a-stark:
## Description

Enhance mainly the saved output of pulsed measurements. The following changes had to be made for that:
- Changes to the file core/util/units.py:
  - Add to the Class ScaledFloat the ability to retrieve the scaled factor.
  - Implement a variety of window functions (together with their correct amplitude scale factors), which can be applied before performing a fourier transformation. This prevents mainly spectral leakage.
  - Rename the method compute_dft to compute_ft and greatly enhance the capabilities of that function (i.e. add abilities to control window functions, base line correction, zeropadding and calculation of power spectral density instead of fourier transform). Apply this rename throughout the whole qudi repository.
- In pulsed_measurement_logic:
  - Replace the fourier transform function my the more general one from core/util/units.py.
  - Restructure the method save_measurement_data, add doc strings, add the possibility to save the fit parameter directly in the graph. Enhance significantly the appearance of the output graph (see added screenshots below).
  - Add variables to control the fourier transform, right know mainly for scripting abilities. These variables are not connected to the GUI so far, hence, you cannot apply those changes through the pulsed main GUI. The reason for that is that the parameter changes of the fourier transform are not handled separately in the logic. They are all bounded to the signal sigSignalDataUpdated. Making the fourier transform parameter accessible by the GUI requires an adaptation of that signal and its chain to the GUI. That should be done in another pull request to prevent breaking changes by this pull request.
- In pulsed main GUI:
  -  Make the display of the signal graph and the fit graph similar to the saved graphs.
  - Use the excising qudi color codes for a better visual representation of measurement data, errors and fit.  

- Please note, 'a.u.' is the abbreviation for 'atomic units', whereas 'arb.u.' is the one for 'arbitrary units'.  

This branch has been rebase onto master.

## Motivation and Context

Reduce hopefully the need for post processing the data upon save.

Moreover the calculation of a fourier transform should be just in one place (and not individually implemented in several files)

## How Has This Been Tested?

The data shown in the 'Screenshot' section are taken during experiment. The same configuration was also tested in the dummy (default) configuration.

## Screenshots:

This was the output graph before the changes:

20170822-1143-18_test-pulse_pulsed_measurement_fig

This pull request enables such output graphs:

### Bare measurement data without fourier transform and fit

20170908-1959-16_test_pulsed_measurement_fig

### Measurement with fit and without fourier transform

20170908-1959-06_test_pulsed_measurement_fig

### Measuremetn with fit and with fourier transform

20170908-1959-33_test_pulsed_measurement_fig

### Increase Zero padding and apply a hamming window

20170908-1717-25_test_pulsed_measurement_fig

### Appearance in the Pulsed GUI:

If the fourier transform parameters are altered in the pulsed_measurement_logic (e.g. by script or by console) the GUI shows the following graphs (where the plot appearance has been adapted to the output graph). Zeropadding set from 0 to 20 (highly insane value) and hamming window was chosen in Fourier transform:

screenshot_zeropad20_hamming_with_new_points

## Types of changes

## Checklist:

### DIFF
--- a/core/util/units.py
+++ b/core/util/units.py
@@ -416,9 +416,37 @@ def in_range(value, lower_limit, upper_limit):
 
 
 def get_ft_windows():
-    pass
+    """ Retrieve the available windows to be applied on signal data before FT.
 
-def compute_ft(x_val, y_val, zeropad_num=0, window='none', base_corr=True, psd=False):
+    @return: dict with keys being the window name and items being again a dict
+             containing the actual function and the normalization factor to
+             calculate correctly the amplitude spectrum in the Fourier Transform
+
+    To find out the amplitude normalization factor check either the scipy
+    implementation on
+        https://github.com/scipy/scipy/blob/v0.15.1/scipy/signal/windows.py#L336
+    or just perform a sum of the window (oscillating parts of the window should
+    be averaged out and constant offset factor will remain):
+        MM=1000000  # choose a big number
+        print(sum(signal.hanning(MM))/MM)
+    """
+
+    win = {'none': {'func': np.ones, 'ampl_norm': 1.0},
+           'hamming': {'func': signal.hamming, 'ampl_norm': 1.0/0.54},
+           'hann': {'func': signal.hann, 'ampl_norm': 1.0/0.5},
+           'blackman': {'func': signal.blackman, 'ampl_norm': 1.0/0.42},
+           'triang': {'func': signal.triang, 'ampl_norm': 1.0/0.5},
+           'flattop': {'func': signal.flattop, 'ampl_norm': 1.0/0.2156},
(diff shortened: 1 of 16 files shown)


Neverhorst:
These changes definetely make sense and improve pulsed measurements. Thank you.
I will check it out on monday to see it in action in the lab.
The measurement data graph is a bit too colorful now in my opinion but thats just a matter of personal taste.

Neverhorst:
I get a warning from matplotlib when saving a measurement:
```
C:\Anaconda3\envs\qudi\lib\site-packages\matplotlib\axes\_axes.py:545: UserWarning: No labelled objects found. Use label='...' kwarg on individual plots.
  warnings.warn("No labelled objects found. "
```

a-stark:
Indeed, there is a label missing, causing that warning.

PR #264, opened 2017-09-28
agilent microwave hardware file updated

SimonSchmitt88:
## Description

Updated the agilent microwave hardware file so that it fulfills the interface and that the mw source works stable

## Motivation and Context

The old file was not up to date and for example ODMR was not working

## How Has This Been Tested?

Extensive tests using for example spectrum analyzer, notebooks and NV-ODMR  (Model N9310A)

## Types of changes

- [] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/hardware/microwave/mw_source_agilent.py
+++ b/hardware/microwave/mw_source_agilent.py
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 
 """
-This file contains the Qudi hardware file to control SMIQ microwave device.
+This file contains the Qudi hardware file to control Agilent microwave device.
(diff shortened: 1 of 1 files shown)


a-stark:
Thank you for the changes.

SimonSchmitt88:
Sorry, I indeed replaced the current file without checking if there were recent changes. (I did not expect someone is working on that hardware file - my fault)

I took all your comments into account and did changes according to them. Since the MWsource is not in Ulm but in Shanghai, I am not able to test these changes with a device

PR #265, opened 2017-09-28
Making confocalgui a bit more convenient

Geegee:
I want to implement a learning curve for the usage of the confocal
in the sense that you will get faster the more you use it (essentially learning shortcuts).
## Description

Commit current changes: - Doubleclick checks action_zoom (Zoom Button)
                                          - zoom is disabled after zooming 
                                          - Added some shortcuts in confocal gui
                                          - In all QuDi :
                                                                        - If QActions have a shortcut + a toolTip then I appended 
                                                                          the shortcut to the toolTip otherwise I created a toolTip
                                                                          containing the key combination
I also wanted to update the shortcuts in the doxygen, but don't know how to update it. Doing 
it automized would be ideal (I have all hotkeys and their actions from parsing xml files, but just don't
know how to access the doxygen).
## Motivation and Context

I changed the zoom to deactivate after usage because I think this is how it is in most other programs.
( and you don't need to reach for the z key when leaning back in your seat :) )

## How Has This Been Tested?

I just played a bit find NV's in the dummy and tested every functionality.

## Types of changes

## Checklist:

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -19,7 +19,7 @@ install:
     - 'conda env remove --yes --name qudi || exit 0'
     - "conda env create -f %APPVEYOR_BUILD_FOLDER%\\tools\\conda-env-win8-qt5.yml"
     - "activate qudi"
-    - "python %APPVEYOR_BUILD_FOLDER%\\tools\\qudikernel.py install"
+    - "python %APPVEYOR_BUILD_FOLDER%\\core\\qudikernel.py install"
 
 # Not a .NET project, we build in the install step instead
 build: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ before_install:
 
 install:
   - python -c "import matplotlib.pyplot" # generate matplotlib font cache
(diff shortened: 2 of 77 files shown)


Geegee:
I changed now the script so only files with changes (new tooltips) are rewritten.

Regarding the change for the double click. First I guess its a matter of preferences how you 
want it to work (only on the confocal or everywhere). Second the double click code is more
or less takem from the python wiki: https://wiki.python.org/moin/PyQt/Distinguishing%20between%20click%20and%20double%20click
Yes there is the mouseDoubleClickEvent but you also have to implement yourself that it
distinguishes between single and double click.

To make it only work in the xy and depth Viewidgets I think one has to overwrite the 
classes like its done now with the main window, but I'm not really sure how to do it 
and don't think I want to put in the effort.

Neverhorst:
Check out the latest changes I introduced. Does that make sense? it's a bit less overhead and works just as you intended.

Geegee:
Thanks looks definitely cleaner and simpler now.

PR #266, opened 2017-09-29
fixes to make magnet working again

SimonSchmitt88:
Magnets should work agaid with QuDi

## Description

The main issue was that 'move_rel' was both a function and a dictionary. So I renamed the function to move_relative. This is just a quick fix.

## Motivation and Context

Otherwise relative movements with Magnet GUI are not working

## How Has This Been Tested?

With this changes the Magnet GUI is working for me again

## Types of changes

## Checklist:

### DIFF
--- a/gui/magnet/magnet_gui.py
+++ b/gui/magnet/magnet_gui.py
@@ -1002,9 +1002,10 @@ class MagnetGui(GUIBase):
 
         movement = dspinbox.value() * direction
 
-        self._magnet_logic.move_rel({axis_label: movement})
+        self._magnet_logic.move_relative({axis_label: movement})
         # if self._interactive_mode:
         #     self.update_pos()
+        return axis_label, direction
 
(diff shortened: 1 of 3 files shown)


a-stark:
Thank you for the fix. Could we do it differently, i.e. renaming the dict move_rel to `move_rel_dict` and leaving the function `move_rel` as it is? In my opinion, this would be less confusing, since the interface method would be equivalent (is also called `move_rel`).

Moreover, I have some more fixed for the magnet logic in the branch [magnet_fixes](https://github.com/Ulm-IQO/qudi/tree/magnet_fixes), which are not finished yet. It might be a good idea to work on that branch. Apart from the fact, that the magnet logic needs desperately an overhaul (together with the gui), I would try to make it in the first place operational.

jochen-scheuer:
As I see from the commit Simon made the changes. Is it possible to pull it to the master? My coding would depend on this...

PR #267, opened 2017-09-29
corrected many treshold to threshold

SimonSchmitt88:
improved threshold extraction method

added a new extraction method which chops a ungated timetrace into a gated one

added a new analysis method without normalization

Threshold was many times spelled incorrectly (treshold)

Fixed pulse extraction threshold method

Added new pulse extraction method

Added new pulsed analysis method

## Description

The fix makes threshold detection quicker. If there are only a certain number of bins below the threshold the laser pulse is not choped into pieces anymore

The new pulse extraction method takes an ungated timetrace and firstly converts it into a gated timetrace . This can be done by knowing the time of the laser pulses in the generated sequence. The gated timetrace is then normalized by gated_conv_deriv-method. Attention this is not working if only GUIs are used yet!

Sometimes it is desired that the data is not normalized. So this method only adds up the photons detected during the signal-bins

## Motivation and Context

spelling fix for readability
threshold_detection fix improves performance
new extraction method improved performance for ungated timetraces
new analysis method increases options

## How Has This Been Tested?

Tested for Fastcomtec using jupyter notebooks...

## Types of changes

## Checklist:

### DIFF
--- a/gui/pulsed/pulsed_extraction_external_gui.py
+++ b/gui/pulsed/pulsed_extraction_external_gui.py
@@ -155,8 +155,8 @@ class PulsedExtractionExternalGui(GUIBase):
         ignore_first=self._mw.ignore_first_CheckBox.checkState()
         method=self._mw.analysis_method_ComboBox.currentText()
         param_dict={}
-        if method == 'treshold':
-            param_dict['count_treshold']=self._mw.treshold_SpinBox.value()
+        if method == 'threshold':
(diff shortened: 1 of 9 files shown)


SimonSchmitt88:
Thanks a lot for your feedback. I removed the 'ungated_gated_conv_deriv' method for now and fixed the threshold and tolerance issue

Neverhorst:
One remark: If you set the threshold too high (no counts can be found) the extraction will throw "index out of bound" errors.

PR #271, opened 2017-10-12
Adding human readable fit result for poissonian double fit

jochen-scheuer:
Adding human readable fit result for poissonian double fit

## Description

Adding human readable fit result for poissonian double fit

## Motivation and Context

With this fix one can call the double poissonian now such that it can be used in the fit 

## How Has This Been Tested?

I tested it in the testing notebook and in ODMR. 

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:


jochen-scheuer:
Can you name a case where mu does not describe an event rate? - Actually I just copied it from the definition of the poissonian fit result

Issue #200, opened 2017-05-03
POI manager delete last position unfunctional


cyz75:
The button 'delete last position' gives an error. 
This function is somehow crucial for working with the POI manager because otherwise one wrong fit in the optimizer leads to loss of all positions.
AttributeError: 'NoneType' object has no attribute 'get_key'
Traceback (most recent call last):

  File "C:\HYPERDIAMOND\Software\qudi\gui\poimanager\poimangui.py", line 650, in goto_poi
    self._poi_manager_logic.go_to_poi(poikey=self._poi_manager_logic.active_poi.get_key())

Delete last position is now a function of the class PoI but I wasn´t able to use it correctly i.e. to delete the last position shift of all POIs

    def delete_last_position(self):  # TODO:Rename to delete_last_position
        """ Delete the last position in the history.

        @return float[4]: the position just deleted.
        """

        if len(self._position_time_trace) > 0:
            return self._position_time_trace.pop()
        else:
            return [-1., -1., -1., -1.]

Neverhorst:
#273 fixed it

PR #273, opened 2017-10-13
Bugfixes in poimanager

jochen-scheuer:
## Description

Fixing bugs in poimanager. 

## Motivation and Context

Some crucial functionality were not working anymore. Many of them are solved by this pull request.
- #200 
- Additionally error messages for not working functionality was included (Autofind/Optimize ROI/Anchor optimization). 
- The reorient ROI was fixed by adjusting the units to µm. 
- Adding "if self.active_poi is None" to move poi/manual update poi/... such that no error message appears in case no poi is selected.
- Fix new ROI button, there was a wrong reference.
- Change estimator of optimization to MLE version. 
- Adding test of 2D gaussian fit to gaussian fit testing notebook.
- Remove 4th channel in confocal_scanner_dummy such that poimanager works with confocal_scanner_dummy 

## How Has This Been Tested?

I checked all functionality in dummy and on a setup2. I checked also the reorient function with a rotated sample and it worked.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)

## Checklist:


jochen-scheuer:
Thanks for the helpful comments. I changed the rotation window to Si units and scientific spin boxes. 

I fully agree that the change in the scanner dummy could lead to problems in other modules. Although I did not encounter any in the dummies I tested.

This 'dummy' fix would not help if the confocal_scanner would actually have more than 3 axes. In this case the POI manager would not work at all. It seems that the confocal_logic has hard-coded 4 channel possibilities ['x', 'y', 'z', 'a']. The poi_manager_logic needs the x,y,z values. In the last commit I changed the _confocal_logic.get_position() to _confocal_logic.get_position()[:3] such that only the first three values are taken which correspond to x,y,z. And the dummy was changed back to  ['x', 'y', 'z', 'a'].

What do you think about it?

PR #282, opened 2017-10-19
Slider behavior intended?

Geegee:
When changing the position with mouse click instead of drag
the value is not updated, also the slider is unresponsive for movements with the arrow keys. 
Changed therefore from sliderReleased signal to update the conv_deriv value in the software to valueChanged. Will this be a problem for the performance of the software, because it changes the 
value to fast? Then need to distinguish between dragging the mouse and changing the value otherwise.
Maybe someone can test it on an working setup.

One liner 
selection_008

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x ] My code follows the code style of this project.
- [ x] I have checked that the change does not contain obvious errors (syntax, indentation, mutable default values).


Neverhorst:
Depending on what the signal triggers these rapid signal fire situations can be extremely CPU heavy.
In that case it basically only changes a value in a logic module so it should be fine... but I would avoid it if it's not neccessary.
Is it neccessary in this case?

Neverhorst:
Ah, damn, how do I change the approval?
Please change also the disconnect signal in the deactivation and adjust it accordingly.

PR #286, opened 2017-10-20
add simple building block functions for sequencing with AWG70k

SimonSchmitt88:
## Description

Added four simple building block function which allow to generate a new sequence, add waveforms, load and make a sequence continous

Additionally the 'get_loaded_asset' method is rewritten so that the actually loaded aset is returned and not a variable

## Motivation and Context

Although there is already more advanced functionality for sequences, this adds a very basic and simple possibility to generate and run sequences

If the asset is changed by hand the old 'get_loaded_asset' method did not return the actual loaded asset

## How Has This Been Tested?

Tested for AWG70001A

## Types of changes

## Checklist:


Neverhorst:
Fixed a bug with the 2-channel AWG70k when calling get_loaded_asset.

PR #290, opened 2017-11-03
APT motor generalised to read hw config from file AND remove code duplication

latchr:
## Description
APT stage class has been generalised so that there is no need for the specific one-, three-, and four-axis stage classes. Now the APTStage class properly reads an arbitrary number of axes as specified in the config file. Constraints are now read from the config file, rather than hard-coded into the aptmotor.py module.

Anyone using APTmotor stages will need to update their config files. Documentation has been updated to include an example config entry in the APTStage class docstring.

## Motivation and Context
This work started as an effort to move the hard-coded constraints out of the aptmotor.py module and into the config. That required some changes to the layout of the aptmotor config parameters in the config file. While working through the process of reading constraints from the config file, it turned out to remove the need for specific one-, three-, four-axis stage classes. This dramatically reduces code duplication in this module, which is a bonus.

## How Has This Been Tested?
This has been tested with a Thorlabs rotation stage. We do not have hardware to test linear stage movement.

## Types of changes

## Checklist:


latchr:
I had a verbal conversation with @a-stark, who mentioned the following comments:

1. The Thorlabs dll filepath is currently hard-coded into the APTmotor class. This could be read from the config file, allowing for the file to stay where Thorlabs puts it. 

2. The setHardwareLimitSwitches command sets a mode where the stage can move forwards and backwards. The TODO comment question needs to be replaced with a better explanation.

3. The _save_pos and _get_magnet_dump have a purpose beyond just the manget application - namely, to save the "current pos" and calculate a correction factor if the stage is power cycled and ends up with a new "home" position.  This feature was "never finished". This is a good idea, and things should be renamed/explained a bit so that it is more transparent.

Neverhorst:
Is it right to assume that the branch is not yet ready to merge into master because the afore mentioned points are currently addressed? (WIP?)

latchr:
Actually, this has really been waiting for someone to double-check it. I've spoken with @a-stark on Skype about this, and it is really ready to merge. The biggest problem is that it will break config for anyone using APT-Motor stages. The new config is not hard to master, and in fact the correct form of the config is shown in the docstring for the class!

@Neverhorst, I recommend that this is merged and that the outstanding details (fixing the names of _get_magnet_dump to be more general) can be a follow-up task.

PR #292, opened 2017-11-16
qdplotter works for two curves

NicolasStaudenmaier:
adapt qdplotter that it can plot two curves and can decide whether to clear old plots

## Motivation and Context

upgrade, can also plot fit
## How Has This Been Tested?

with dummy and jupyter notebook

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] I have checked that the change does not contain obvious errors (syntax, indentation, mutable default values).
- [ x] I have tested my changes using 'Load all modules' on the default dummy configuration with my changes included.

### DIFF
--- a/gui/qdplotter/qdplottergui.py
+++ b/gui/qdplotter/qdplottergui.py
@@ -143,7 +143,7 @@ class QdplotterGui(GUIBase):
 
         self.curves = []
         pen_colors = cycle(['b', 'y', 'm', 'g'])
-        for ii in range(np.shape(self._qdplot_logic.indep_vals)[0]):
+        for ii in range(len(self._qdplot_logic.indep_vals)):
             self.curves.append(self._pw.plot())
             self.curves[ii].setPen(next(pen_colors))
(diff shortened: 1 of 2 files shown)


a-stark:
Hi @NicolasStaudenmaier, thanks for the pull request. 
Do not forget to assign reviewer(s) for the pull request, then it might be processing faster.
I will not be able to test this pull request this month, but maybe @latchr might have a quick look on it.
At first glance, it looks ok to me. 
I would probably go directly for the multi dimensional approach, so that `self.indep_vals` and `self.depen_vals` are always saved as a multi dimensional array (this would save you several if statement checks).
In the method `set_data`, it would not harm to cast the input always to a numpy array (self.indep_vals = np.array(x) ... and skip the list checking).
Then you can use the much shorter and more intuitive np routine for min and max, like:
domain_min = self.indep_vals.min(axis=1)
domain_max = self.indep_vals.max(axis=1)

NicolasStaudenmaier:
Thanks for the suggestions. I have edited the code and now it also works for multidimensional plots.

PR #297, opened 2017-12-23
Simplifying Modules

tobiasgehring:
This patch simplifies modules by nesting the fysom state machine instead of subclassing it.

## Description
Instead of subclassing fysom it is now nested and the state machine is accessible by the attribute module_state. To get the current state you can either use module_state.current (Fysom standard) or module_state as callable: module_state().
Some methods in the module base class like get_home_dir and get_main_dir have been moved to core.util.modules, others have been deprecated and assigned a warning.

## Motivation and Context
The fysom state machine includes many dynamically added attributes and methods and it happens quite often that one overrides a methods without noticing. Examples are run, stop, trigger, lock and unlock.

## How Has This Been Tested?
I started qudi and a couple of modules. However this pull request includes changes to many modules so I would appreciate if someone can test some more modules. I haven't used "Load all modules" on default configuration yet since for some reason my python installation causes qudi to raise errors (pandas has some issues).

## Types of changes

## Checklist:

### DIFF
--- a/core/manager.py
+++ b/core/manager.py
@@ -924,7 +924,7 @@ class Manager(QtCore.QObject):
         if not self.isModuleLoaded(base, name):
             logger.error('{0} module {1} not loaded.'.format(base, name))
             return False
-        return self.tree['loaded'][base][name].getState() in ('idle', 'running', 'locked')
+        return self.tree['loaded'][base][name].module_state() in ('idle', 'running', 'locked')
 
     def findBase(self, name):
         """ Find base for a given module name.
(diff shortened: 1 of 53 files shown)


drogenlied:
I have no issues with this on my Debian PC at home.

Can anybody comment if their custom notebooks use `getState`?

Neverhorst:
Yes, my notebooks for pulsed experiments use getState. But that shouldn't be a problem to change.

tobiasgehring:
could you approve the pull request?

tobiasgehring:
thanks. Unfortunately I'm not authorized to merge it. I believe this is because I'm outside collaborator and the master branch is protected.

Neverhorst:
Hey I'm back from the holidays now.
Looks good also from my side, thank you. I will merge it. 
Could you please write an email over the announce list with the important changes that could break peoples scripts etc. (like using `getState`)?

PR #304, opened 2018-01-22
 Fix to PulseStreamer pulse generation. 

bens0n:
Eliminates cases where the final pulse in a sequence isn't propagated into the sequence uploaded to hardware

## Description
Removes bug in the sequence generation logic for the Swabian Instruments Pulse Streamer. The raw sequence is compressed into pulses with a given number of channels active and a duration. The previous logic was faulty and could accidentally eliminate a pulse from the end of the sequence. 

## Motivation and Context
This change ensures that the sequence generated is true to the desired sequence.

## How Has This Been Tested?
Tested by generating and inspecting a large number of sequences to check for edge cases.

## Types of changes

## Checklist:

### DIFF
--- a/logic/samples_write_methods.py
+++ b/logic/samples_write_methods.py
@@ -365,6 +365,39 @@ class SamplesWriteMethods():
         If both flags (is_first_chunk, is_last_chunk) are set to TRUE it means
         that the whole ensemble is written as a whole in one big chunk.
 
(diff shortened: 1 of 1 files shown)


bens0n:
I've written up a docstring on the design and implementation of the sample write function.

Neverhorst:
Thank you very much. Thats just what I had in mind.

Issue #221, opened 2017-07-08
"Load all modules" doesn't change "Load MODNAME" into "Show MODNAME"


tobiasgehring:
Usually if one clicks "Load MODULENAME" in the GUI tab to load a module, the caption of the button is changed into "Show MODULENAME".

This does not happen when a user clicks "Load all modules".

Expected behavior: Caption is changed to "Show MODULENAME" in the GUI tab.

PR #306, opened 2018-01-29
improved visual indication for unloaded modules

mvbnano:
## Description
This update deactivate the module reload button in `managergui.py` if the module is not yet loaded. The button reactivates once the module has been loaded.

## Motivation and Context
This was initiated in the teaching labs where inexperienced users were unsure whether modules had been loaded. They requested additional visual feedback, see issue QMAPP-mq/qudi#23

## How Has This Been Tested?
This has been tested by loading modules in the dummy configuration.

## Screenshots:

image

## Types of changes

## Checklist:

### DIFF
--- a/gui/manager/managergui.py
+++ b/gui/manager/managergui.py
@@ -620,6 +620,9 @@ class ModuleListItem(QtWidgets.QFrame):
         """ Send signal to load and activate this module.
         """
         self.sigLoadThis.emit(self.base, self.name)
+        
+        # Instant return to checked to prevent visual lag before checkModuleState completes
+        self.loadButton.setChecked(True)
 
     def reloadButtonClicked(self):
         """ Send signal to reload this module.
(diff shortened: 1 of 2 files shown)


latchr:
Actually, I think if the reload button is deactivated then it makes things more consistent for the "deactivate" and "clean up" buttons to also be deactivated if the module is not loaded. @mvbnano, see if you can add this. @drogenlied, the only problem might be if a user needs to "clean up" a module that is not loaded - does this happen?

mvbnano:
Commit 4310e7c makes the changes requested by @latchr 

The three cases are outlined below:

- If the module is not loaded:
    - [ ] reload
    - [ ] deactivate
    - [ ] cleanup

- If the module is idle:
    - [x] reload
    - [x] deactivate
    - [x] cleanup

- If the module has encountered an exception:
    - [ ] reload
    - [x] deactivate
    - [x] cleanup

The relevant segment of code is shown below.

```python
    def checkModuleState(self):
        """ Get the state of this module and display it in the statusLabel
        """
        state = ''
        if self.statusLabel.text() != 'exception, cannot get state':
            try:
                if (self.base in self.manager.tree['loaded']
                        and self.name in self.manager.tree['loaded'][self.base]):
                    state = self.manager.tree['loaded'][self.base][self.name].module_state()
                    self.reloadButton.setEnabled(True)
                    self.deactivateButton.setEnabled(True)
                    self.cleanupButton.setEnabled(True)
                else:
                    state = 'not loaded'
                    self.reloadButton.setEnabled(False)
                    self.deactivateButton.setEnabled(False)
                    self.cleanupButton.setEnabled(False)
            except:
                state = 'exception, cannot get state'
                self.reloadButton.setEnabled(False)
                self.deactivateButton.setEnabled(True)
                self.cleanupButton.setEnabled(True)

            self.statusLabel.setText(state)
```

drogenlied:
Reloading a remote module where the state cannot be obtained because the remote side has been closed is the biggest case I would see for keeping reload active.

I am not totally sure if the backend code will do the right thing right now when reloading a remote module, I will have to look into that.

latchr:
I like the idea of having the "cleanup" button disabled when the module is loaded. 

However, how does a user unload a module? While there is a "deactivate" button, it seems this is not the same thing as "unloading" the module. Does the deactivate button simply run the `on_deactivate()` method of the module but keep it "loaded"?  Does this mean that the correct procedure for "cleanup" is to first restart Qudi, then press cleanup before loading the module?

drogenlied:
Unloading a python module is not really possible and Qudi reflects that.

With regards to cleanup, the `deactivated` and `not loaded` states are equivalent, so it is OK to do cleanup when deactivated.

latchr:
I have improved this in a few different directions.

1. The "Load <module>" button is now checkable, and is set to remain checked once the module has been loaded. The module cannot be unloaded, and so this button stays permanently checked after the first click.

2. The "Load <module>" button has its text changed for all modules. The GUI modules turn into "Show <module>" as before, but all other modules turn into "<module>" (without the word "load"). This is much clearer for a user, as it no longer implies a function (load) that is unavailable.

3. When  a module has status "deactivated", the load button has text changed to "Activate <module>".

4. In the process of doing the previous modification, I fixed a small problem with the position of the button setText code. It was previously in the loadButtonClicked function, but this made it not responsive to the module being loaded by other means. For example, I loaded the counter gui from the console using two lines
    `s = man.modlist[0]`
    `s.sigLoadThis.emit(s.base, s.name)`
    and the manager widget did not update to "Show Counter".  Now this is fixed because the button text is re-written in the checkModuleState method.

5. Cleanup is now enabled only when the module is not loaded. When the module is loaded the "reload" and "deactivate" buttons are enabled.

It now looks very nice!

latchr:
module_gui

latchr:
Fixed #221 by merging this request.

PR #315, opened 2018-02-02
ScienDSpinBox alternative

Neverhorst:
This branch introduces a new implementation of the scientific DoubleSpinBox widget.

## Description
This is a proposal to replace the ScienDSpinBox class which subclassed QAbstractSpinBox with a new one that subclasses QDoubleSpinBox to reach a more uniform behavior of different kinds of SpinBoxes.

Please check if your most used GUIs are working as intended with this branch.

Also if you have any input for different behavior/features please let me know.

## Motivation and Context
The existing implementation of ScienDSpinBox is in principle working. 
The problem is that the usage and the behavior differs quite a lot from the standard Qt QDoubleSpinBox.
These issues range from non-standard widget geometry behavior over non-standard event handling to a non-standard programming interface (see e.g. `setOpts()`).
These issues are causing some headache in non-trivial GUI designs (e.g. QTableView item delegates).
In order to make the ScienDSpinBox more similar to the "standard" SpinBoxes I replaced the class to inherit QDoubleSpinBox instead of QAbstractSpinBox. This has the advantage that you only need to reimplement functions that are related to the scientific notation and other non-standard functionality.
The main work was done by writing a custom regex based validator (QValidator).

**Features added to QDoubleSpinBox:**
- supports user input in scientific notation (i.e. `42.12345e-8`) and/or with SI-unit prefix 
(i.e. `42e3 M`, `42 G`)
- Step size (e.g. by mouse wheel or arrow up/down) is dynamically adapted to the second most significant digit of the integer part of the number. This is the default behavior and is enabled by setting stepsize using `setSingleStep(stepsize)` for `stepsize <= 0.0`. Setting `stepsize > 0.0` will result in fixed absolute steps. The precision is adapted in order to display the step size (see next bullet). 
- The number of fractional digits to show is dynamically derived from the number of digits the user has typed into the ScienDSpinBox. This number of digits is maintained until the user specifically enters a different number of digits. Default is 2 fractional digits.

**Overwritten methods of QDoubleSpinBox:**
- `setSingleStep`
- `validate`
- `fixup`
- `setPrefix`
- `setSuffix`
- `valueFromText`
- `textFromValue`
- `stepBy`

## How Has This Been Tested?
Tested on Win 8.1 64bit with default installation of the qudi Python environment for Windows.
Played around with the ScienDSpinBoxes of ConfocalGui and PulsedMainGui.
Also loaded all modules with default.cfg in order to check every GUI for non-functional SpinBoxes.

## Types of changes

## Checklist:


Neverhorst:
@kay-jahnke, could it be that you checked out an early version of the branch? The lines of code throwing the error (e.g. the import of `ScienSpinBox` in line 46 of pulsed_maingui.py) are not present anymore. See "Files changed" tab in this pull request.

Neverhorst:
Regarding the suggested changes. I'm not sure how to bring the QtDesigner to not overwrite the default settings in `__init__` of ScienDSpinBox.
If you create an UI file with designer it will set decimals to 2 and singleStep to 1.0.
One could either write a C++ plugin for designer or overwrite `setDecimals` to always use max. But the latter case wouldn't solve the singleStep problem.

kay-jahnke:
@Neverhorst, thanks for the note:
Jup I forgot the update, so now the errors were resolved. 
I do however get another error, that relates to the module "odmr_counter_microwave_interfuse" which is caused by the last merge. So I will open an individual issue for that.

The issue with the QtDesigner is annoying and it would be good if we found a solution for it. As it stands, it does not hinder the merge of this pull request.

Neverhorst:
Ok, found a workaround:
It's now:
- `setSingleStep(self, val, dynamic_stepping=True)`
- `setDecimals(self, digits, dynamic_precision=True)`

The additional optional flag parameters are now used to indicate if the dynamic stepping/precision should be used or not. By default the dynamic behaviour is enabled and the QtDesigner won't change that. If you want to turn this off you need to call these methods in your GUI module with the flags set to `False`.
The magnet GUI is currently an example for fixed absolute step sizes.

kay-jahnke:
Thanks for the Update. It works and makes the code so much cleaner.

I still get the ODMR-Interfuse-Error, but that will vanish, when you merge the trunk. So in my opinion it's ready to go.

a-stark:
thanks @Neverhorst for the implementation. Indeed, the geometric behaviour was really inconvenient. Let me just test this implementation on Win10 since I encountered there some displaying issues with the AbstractSpinBox implementation. I will come back to you soon.

Neverhorst:
Thanks @a-stark for looking over the changes as well. 
Since it basically affects all GUI modules a third pair of eyes certainly helps.

a-stark:
In my opinion, the SpinBox Widget is a very important implementation, therefore we I am here more careful in the review as it affects nearly all display items. I tested the new spinbox implementation thoroughly and the string representation and conversion is working as expected. However, I have some concerns with the current implementation (which is mainly related to the stepping procedure):

1. The current stepping is actually not a real logarithmic stepping, it is just a relative stepping (i.e. relative to the current SI prefix). You can see that for instance in the ODMR GUI with the 'Step' SpinBox, try to decrease the value and you will reach 0 before it changes say from MHz to kHz. This was for me the most useful feature, which I thought a SpinBox should have. 

2. The single step (with dynamic_stepping=True) increments the value before the point, which I find unconventional. I would assume that the normal increment would (slowly) increment the current value, which would be the case if the number after the point is incremented. By pressing the shift key I would expect that the value before the point is incremented. (This is a rather quick fix, so no big deal).

3. There are use-cases, where a minimal stepsize needs to be defined, but if it is not reached, then the logarithmic step size is used. This is the case for instance in the Magnet GUI, you will step down logarithmically until you reach the limitation of your stepper motor, which is denoted by the minimal stepsize. So a transition from logarithmic to relative and then to absolute stepping is missing, which I find to be a necessary requirement. (Otherwise, a step in the GUI would not correspond to an actual step in the hardware causing a misleading interpretation. Or if the hardware does not permit a smaller step, then the stepping becomes not applicable in the spinbox and the value cannot be changed by stepBy method. Both cases are not desired in my opinion).

4. I am missing the restore functionality, i.e. if you type in a value, but then you decide during the typing process to keep the old value (by pressing ESC). 

5. There is also another subtle implementation detail, which might cause accumulation of rounding errors (which should be actually prevented according to the docstring in the stepBy method). For this reason, the [decimal](https://docs.python.org/3.7/library/decimal.html) module was used in the SpinBox implementation which offers several advantages over float (due to that it was also used in the SpinBox implementation of pyqtgraph).

6. I checked the geometrical issues I had with the SpinBoxes and it turned out that it was more a poor (i.e. static) ui design and not really related to the spinboxes. Could you provide me with examples, where you have encountered problems in the geometrical display?

7. To my mind, I think that the current SpinBox reimplements all required QDoubleSpinBox methods which we use. Do you miss here any method? Or are there some strange behavior with the current spinbox I am not aware of? Maybe it would be easier to fix the current SpinBox if there are some issues.
The setOpts method is in my opinion actually pretty handy, since you can set a lot of features at once, but you do not have to use it, as all the option settings are done according to QDoubleSpinBox implementation.

Given the points above, I have the impression that it is easier to fix the problems with the geometrical representation and the delegate issues. 
Another suggestion from my side would be to use this implementation just in the delegates of the QTableView and to fix the issues of the SpinBox classes in another pull request. (Could you give me some examples, where you encounter problems in the delegate. I could help fixing it.)

a-stark:
I add the fixes to the ODMR GUI ui which was rather statically implemented. This caused issues on Windows 10, Ubuntu and especially on 2K and 4K displays, where the OS scales the font size caused and the SpinBoxes in the ODMR gui become unreadable. Another nice feature: The "ODMR Fit" and the "ODMR Control Parameters" Dockwidgets become adjustable.

a-stark:
That looks great. Thumbs up for the usage of private instant variable (double underscore), I would also go for that! Let me know, if you need some help in the implementation process. 
Just in case, here is the c++ implementation of the qspinbox (both, the integer and the double version), where the actual implementation of the added methods (in comparison to qabstractspinbox) could be seen:
https://github.com/qt/qtbase/blob/5.11/src/widgets/widgets/qspinbox.cpp

Neverhorst:
Thanks I already found the C++ implementation but it was not helping that much I'm afraid.
Currently I'm trying to avoid a mismatch in displayed value and actual value upon editingFinished. This happens because the scaled si number (between 0 and 999.99...) has a different machine representation than the number in the actual order of magnitude. This effect is currently present when displaying a number of digits beyond float precision. The spinbox should always show the "real" value upon editingFinished. 
Additionally the pythonic way of using properties is unfortunately incompatible with the Qt naming convention due to the same name for the getter method and the property itself. Thus I did not use it.
Thanks for the offer to help with the implementation but I think in the current state that could lead to some confusion. I would appreciate however if you could comment on the code once it's in a "release" state and point out stuff to improve.

Neverhorst:
Can somebody please check out this branch now and see if everything works as intended? This spinbox should now have quite some advantages over the previous implementation.
The handling of Decimals is now actually making a difference and upon editingFinished they will now also show the exact float representation of the underlying value if you have enough decimals shown.
Furthermore the dynamic precision now makes it possible to invoke decimal precision on the fly by user input.
Also the usually rather buggy Qt implementation of prefix/suffix is now better.
The integer version (ScienSpinBox) is not used in any module yet.

Neverhorst:
Good point @kay-jahnke. I fixed that by setting the cursor position of the underlying QLineEdit to 0 when editingFinished is emitted.

kay-jahnke:
Much better, but I'm still not completely satisfied:
Now, if you leave the focus it shows the first numbers. 
But if the Spinbox has the focus and a lot of digits, you can scroll the mouse wheel to change the value (first digit after the comma). In this case you can change the value without being able to see it.

I think this comes from the fact, that while you are scrolling, the cursor is automatically set to the end.

Neverhorst:
I'm quite annoyed that stuff like that is not the default behavior of a QAbstractSpinBox since it's independent of the exact content. 
In general the Qt SpinBoxes are strange when it comes to details.
But ok, if we implement our own we can as well make it better.
Have a look at the changes, @kay-jahnke.

kay-jahnke:
This is really annoying behaviour of QT.
Now it seems to work fine for the dynamic case.

There is another weird thing however:
odmr-load
On loading the GUI, is now seems to centre (maybe?) the values in the spinbox. If you click the spinbox it than corrects it. Still it would be nice if it could be done correctly from the beginning.

Neverhorst:
What about now?

Neverhorst:
So... everyone happy with merging this? Any objections @a-stark?

a-stark:
Thanks for the huge work concerning this very important widget.
I am afraid that the spin box requires one additional feature (but which can be potentially addressed in a further pull request, if you feel that we should merge the current version).
I remembered that I had actually changed the validator function in the [spinbox file](https://github.com/Ulm-IQO/qudi/blob/utilize_user_data_in_fit/qtwidgets/pyqtgraphmod/SpinBox.py) (the old one) in the branch [utilize_user_data_in_fit](https://github.com/Ulm-IQO/qudi/tree/utilize_user_data_in_fit) to equip the spinbox with additional functionality, to be able to use it in the fit guis. There, the additional values `inf`, `-inf`, `NaN` or `nan` or `NAN` would be necessary. 
If these functionalities are implemented, then I could adjust the branch according to those and make it ready for a pull request.

Neverhorst:
@a-stark, I have seen that `inf` and `nan` are getting passed to setValue but I see a fundamental problem there.
These values have no defined numeric behavior when it comes to spinboxes. What would be the expected behavior for `stepBy`? I mean for `inf` this could be handled but `nan`? What is the purpose of a spinbox holding "not a number"?
I can put handling for `inf` and `-inf` in there by disabling stepping in that case. And the validator is easily   adjusted to that as well.
But it would be good if you could elaborate why a numeric spinbox should hold a non-valid number.

a-stark:
These are valid concerns. 
The main reason is at first compatibility to the [fitgui](https://github.com/Ulm-IQO/qudi/blob/master/gui/fitsettings.py#L506) (as the numeric result of the fit algorithm can take the values `inf`, `-inf` or `nan`). 
Setting a value to `inf` or `-inf` would mean to have no limitation for the value (this is used in lmfit). This is very often used for a range definition, in which a parameter should be varied (here a range is defined by min_val and max_val and both values are displayed via the spinbox).
I would suggest the mathematical behaviour for the stepBy function for the values `inf`, `-inf`, i.e. 'inf+step_val' remains 'inf' (same for '-inf').
The value `nan` is the result of a parameter calculation indicating that something went wrong (e.g. `1/0 => nan`). So `nan` is really used for a displaying behaviour.
I have implemented for 'nan' value the `stepBy` like:
If NaN is present in the Spinbox, then upon one step the value should start from the bounds depending on in which direction the step is performed (step + starts from minimum, step - starts from maximum). By this the step functionality is preserved even if NaN is present in the Spinbox.

Neverhorst:
Yes, I see your point.
I'm just not comfortable with having a spinbox return `nan` when you call value() since this will potentially crash a lot of things. `inf` is compatible with most `float` operations but `nan` is not.
When I understand you correctly, @a-stark, `nan` is only needed for displaying purposes, right? In other words to inform the user of a failed math operation. Why can't we just handle a failed operation like any other error or warning? Otherwise this failed value will propagate through a lot of things and just needs to be handled at a different place.

a-stark:
You are right that it might cause errors, but NaN is actually part of [IEEE_754](https://en.wikipedia.org/wiki/IEEE_754), which defines the standards for floating-point arithmetic. Therefore I believe a spinbox should adhere to this specification. What do you think?

Neverhorst:
Yes, it is part of IEEE_754 but anyways it makes it necessary to explicitly check for NaN every time something receives a value from a spinbox since many functions that accept floats as input will throw an exception with NaN.
In my opinion it would be better to rely on a spinbox to return a number that can be treated as one in every way (so also `inf` would be arguable). 
Having a spinbox that just silently accepts and passes NaN will just distribute the problem to everything else. 
At some point the code HAS to handle NaN in a special way (i.e. throwing an exception that the math operation has failed). Why not forcing the programmer to do sanity checks before setting a spinbox value?

Neverhorst:
The QDoubleSpinBox "solves" the problem by accepting `setValue(float('nan'))` but if you afterwards call `value()` it will return `maximum()`.
Not perfect either but at least it won't return NaN.

a-stark:
I agree that a computation, which produces NaN should raise an exception.
I think the acceptance of a NaN value and the return of a float value might be a good compromise. Let's do this.

Neverhorst:
Ok, @a-stark, I think this is a good solution now.
All these changes are only done in the ScienDSpinBox (not the integer spinbox).

Since the changing of the value to `maximum()` when setting a NaN seems kind of arbitrary to me, I now made the spinbox accept NaN in `setValue` but not change the underlying value. Setting it to some other value might even cause dangerous situations for our hardware.
I introduced however a new flag property `is_valid` that will indicate if the currently available value is valid or might be invalid. When trying to set NaN as value this flag will go to `False` indicating that calls to `value()` might return an invalid float (NOT NaN but the last valid value).
In the GUI this will result in a red frame drawn around the spinbox.
When the user enters a valid text in the spinbox, triggers step up/down or calls `setValue` with a valid number, this flag will go back to `True` and the red frame will disappear.

a-stark:
That looks good! I like the red box. And it makes definitely sense, that an integer cannot hold `-inf` or `inf` numbers.
I think we can truly say that our spinboxes are now not 'on steroids' but 'on speed'.

drogenlied:
One minor thing, maybe this happens only on X11/Linux, but triple-clicking the spinbox to select all contents and then entering a number and a SI prefix deselects everything after the first key is pressed and does nothing afterwards with the cursor sitting between the prefix and the unit.

Neverhorst:
Thank, @drogenlied, for pointing that out.
That is actually the programmed behavior but it was implemented with the cursor position in mind and not a selection (which sets the cursor at the end of the selection).
I did not pay a lot of attention when it comes to selections that select more than the editable part (what you get with `cleanText()`, no prefix/suffix).

When you selected all visible text, the cursor will be after the suffix. Upon a key stroke this cursor position will be corrected to the editable range between prefix and suffix. In your case that should have been after the SI-prefix/before the suffix (at least that was what I could reproduce on windows).
The fact that you could not enter text afterwards was because the validator does not allow characters after the SI-prefix.

I made some changes to `keyPressEvent` so that the case of a selection is corrected differently from the case of an out of range cursor. Now also partial selections that include part of the prefix/suffix are handled properly.
Please check if that works as intended.

drogenlied:
Thanks, that works beautifully.
I don't know if it is related, but now the values that are entered in ODMR/Confocal GUI do not affect anything.

Neverhorst:
Ok, strange... I have not realized until now. Apparently this was caused by `editingFinished()` not being emited upon focus loss by clicking somewhere.
I think the last commit should fix that.

PR #317, opened 2018-02-07
Consistently set current position as an np.array internally & expose as a list through get_scanner_position

cjstephen:
Fixes a bug with refocusing/optimization where returning an ndarray
is incompatible with emitting a refocusing signal. Tested on Warwick
forked system (minimal change prepared against Ulm master)

## Types of changes

## Checklist:


a-stark:
Thanks for that, this fix looks reasonable to me. Could someone test this quickly with a connected hardware?

PR #319, opened 2018-02-12
Units of threshold extraction method in seconds

NicolasStaudenmaier:
The units for the minimum laser length and threshold tolerance of the extraction methods are changed from bins to seconds. 
## Motivation and Context
It is clearer to insert the pulse length than the number of bins.

## How Has This Been Tested?
Has only been tested with Dummy. Worked well.

## Types of changes

## Checklist:


kay-jahnke:
I think this change makes sense.

Could you please still print the bin values to log.debug, just to make it easier for people if they are trying to find there errors.

NicolasStaudenmaier:
Thanks Kay for the suggestion. I have added a print of the bin values to log.debug. This is done each time when the laser pulse is extracted. The other possibility would be to print them when the measurement data is saved. I don't know which solution is the better one.

PR #330, opened 2018-03-06
Updated changelog for release v0.9

Neverhorst:
## Description
updates `changelog.md` to include latest release v0.9
I tried to reconstruct the mayor changes since almost a year. 
For the future we should add new features/changes directly into `changelog.md` when merging into master.

## Checklist:


Neverhorst:
Yep, forgot those. 
But since tests still fail (not sure if merged changes should actually fix this) do we want to include it now in the changelog?

Neverhorst:
Well it seems to me that the errors occur after executing the shutdown.ipynb. (Travis CI and AppVeyor)
Can't reproduce this crash by just executing the notebooks on my PC in the same order so I guess it has something to do with the test environment?

If nobody has an immediate idea of how we can quickly fix that, I suggest we exclude this topic from v0.9  (as it is right now) and merge this changelog.

drogenlied:
Crashing on exit does not matter for the test suite.

Sine and sine exponential fit notebooks are still as fucked as ever though and fail randomly with a rather large chance.

Neverhorst:
Ok, I looked a bit closer and I see what you mean. 
Seems as if all errors are caused by not handling a `stderr == 0` returned by `lmfit` when dividing through it (`ZeroDivisionError`) in `sinemethods.py`. So apart from being a bit crude and intransparent the notebooks are not the problem.
I will introduce a quick fix in this branch and then the tests will hopefully succeed.

But again I must say that the whole concept of the fit methods is extremely messed up. The overhaul to avoid code duplication was the biggest mistake we did. The result is unreadable code which makes it extremely hard for someone to add new fits and complicates debugging. Also it adds as much overhead as it avoids code duplication which renders the whole idea absurd.
Also the concept is not thought trough since there is not even a clear definition on how to compose fit models. What defines an atomic model to build from?

The idea of qudi is that scientists can use/understand this code and expand it. So a readable/simple definition of a fit method must always take precedence over a couple of duplicated lines.

Neverhorst:
The `ZeroDivisionError` in sine fits has been fixed but I'm pretty sure there are some (less frequent) random fails in lorentzian fits.

In order to produce these random errors a bit more reliable I made the `fit_testing_*` notebooks repeat each test 100 times. For that I also had to increase the timeout for cell execution with nbconvert to 10 min (default was 30 sec).

However I'm still a but confused about the unclean exit. AppVeyor seems to always fail due to these last lines in the console:
```
[00:08:43] [NbConvertApp] Converting notebook notebooks/shutdown.ipynb to html
[00:08:43] [NbConvertApp] Executing notebook with kernel: qudi
[00:08:46] [NbConvertApp] Writing 249051 bytes to notebooks\shutdown.html
[00:09:01] Scanning Device will be reset.
[00:09:03] Used Qt API: PyQt5
[00:09:06]       0 [main] sh 1816 cygwin_exception::open_stackdumpfile: Dumping stack trace to sh.exe.stackdump
[00:09:06] Command exited with code 35584
```

@drogenlied: are you sure that this is not causing the check to fail? As far as I understand it now this is unrelated to counting errors in the qudi log.

drogenlied:
There are a few things that are seen as test failures:

 * stderr output in notebooks, increases failure count and test exit code
 * qudi process not running directly after notebook execution, likely a crash
 * qudi process not running 10s after starting it
 * qudi process running x seconds after shutdown
 * errors written in the qudi log file

Ultimately, test success is measured by the test script exit code, so if anything manages to take down the bash/shell process executing the script, that counts as a failure.

Neverhorst:
Ok thank you @drogenlied.
After the little fixes I made, the sh.exe crash seems to be the main remaining reason for the appveyor test to fail.

PR #332, opened 2018-03-08
fixed poi-manager: a feature that was more a bug

kay-jahnke:
The PoiManagerLogic was listening for a signal, that the confocal position was changed and then deleting the active poi (if it wasn't done by the optimizer.

This let to the effect, that you could simply loose your POI by changing the focus between windows.

As far as I know, there wasn't any usecase, where you needed to forget your active POI when changing the position, so this should not hurt anyone.

## Motivation and Context

When changing focus between POI-Manager and Confocal the active POI was lost. This led to some of the scripted measurements crashing.

## How Has This Been Tested?
Tested with dummy, since it is not dependent on hardware (just logic involved).

## Types of changes

## Checklist:

### DIFF
--- a/logic/poi_manager_logic.py
+++ b/logic/poi_manager_logic.py
@@ -256,8 +256,7 @@ class PoiManagerLogic(GenericLogic):
         """ Deactivate the active POI if the confocal microscope scanner position is
         moved by anything other than the optimizer
         """
-        if tag != 'optimizer':
-            self._deactivate_poi()
+        pass
 
     def testing(self):
         """ Debug function for testing. """


drogenlied:
There was some reason for that, but something is clearly broken rignt now.
I think the behaviour was intended for actual manual movement of the cursor, but the assumption "optimizer" == not manual movement clearly is not valid any more and perhaps never was.

Changing window focus and sizes should never affect anything, so that also needs to be investigated.

kay-jahnke:
@drogenlied I still don't see why you want to delete the active POI, even if you move your cursor manually. In my opinion I would actually expect explicitly, not loose my active POI, to be able to come back to it, even after I played around in confocal.

Concerning the general problem of a signal being send, when you change the focus: I think this is as intended. If you had the focus somewhere in a field, that contains a position and now the focus of this field is lost, than the changed value is accepted and an update is sent. This update is also sent, even if your cursor was sitting there without making any actual change.

drogenlied:
I guess previously, the POIU was not deleted, just deactivated, judging by the name of the function.
Deletion is the new and buggy behaviour, and is definitely wrong.
So just merge this, I will probably need to look at some of the leftover "construction sites" soon.

PR #334, opened 2018-03-14
Pulsed: when sigStopMeasurement got None in parameter Python crashes

kay-jahnke:
This was a grave error in the pulsed signal chain:

The signal sigStopMeasurement in PulseMasterLogic has one parameter. This parameter is determined by another function call, normally from the pulsed main GUI. The GUI performs the function call without any parameter and things run smoothly.
When you now however call PulseMasterLogic.stop_measurement(None) from a script, this None parameter gets handed to a signal causing Python to crash miserably. it seems you cannot hand over None as a parameter when emitting a signal.

Fixed it by checking for None in the parameter handed over and changing it to an empty string.

## Motivation and Context
Catch Python unconventional usage of pulsed when running measurements from script.

## How Has This Been Tested?
Just concerning logic and its connection to GUI.
Tested on Dummy.

## Types of changes

## Checklist:

### DIFF
--- a/logic/pulsed_master_logic.py
+++ b/logic/pulsed_master_logic.py
@@ -609,6 +609,8 @@ class PulsedMasterLogic(GenericLogic):
 
         @return:
         """
+        if stash_raw_data_tag is None:
+            stash_raw_data_tag = ''
         self.sigStopMeasurement.emit(stash_raw_data_tag)
         return
 
--- a/logic/pulsed_master_logic.py
+++ b/logic/pulsed_master_logic.py
@@ -609,7 +609,9 @@ class PulsedMasterLogic(GenericLogic):
 
         @return:
         """
-        if stash_raw_data_tag is None:
+        if not isinstance(stash_raw_data_tag, str):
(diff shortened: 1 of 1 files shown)


Neverhorst:
Actually it's not even a bug but just plain wrong usage (by your script probably).
`None` is of a different type than `str` which is defined in the signal signature. Handing different objects to the signal emit always causes these problems. That's exactly why the method `stop_measurement` has an empty string as default parameter defined. So if you want no save tag, it is simply wrong usage to pass None... just pass an empty string or nothing at all.

kay-jahnke:
@Neverhorst I agree with you on the unconventional (you could call it incorrect) usage.
I think the confusion arose, since further down the line, this parameter is checked for an empty string and than converted into None again (in stop_pulsed_measurement of pulsed_measurement_logic).

As the parameter of a function call is directly passed to the signal, if can lead to problems. So my check makes it a bit more robust, especially as there is not only an error, but Python crashes when used incorrectly.
This is why I still see the relevance of this pull request.

kay-jahnke:
Good point actually. 
I will make it even more general and also print a warning.

PR #341, opened 2018-03-20
Zero value scientific SpinBox fix

Neverhorst:
## Description
It was not possible to enter zero values in the LineEdit of the scientific spinboxes.

## Motivation and Context
This bug was caused by a faulty sanity check that triggered to return early when the value to set was zero.

## How Has This Been Tested?
Dummy config

## Types of changes

## Checklist:

### DIFF
--- a/qtwidgets/scientific_spinbox.py
+++ b/qtwidgets/scientific_spinbox.py
@@ -306,7 +306,7 @@ class ScienDSpinBox(QtWidgets.QAbstractSpinBox):
         """
         text = self.cleanText()
         value = self.valueFromText(text)
-        if not value:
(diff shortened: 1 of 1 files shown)


Neverhorst:
@kay-jahnke, this is not quite as you describe it. You can enter everything. However there is a little thing about the validator that prevents entering an SI-prefix in front of a trailing whitespace. So `'123.456 n'` works but `'123.456n '` is not valid. 
If you single click inside the TableWidget, you can edit everything. Just try to enter the SI-prefix after the already present whitespace.
This can be fixed but I have to look more closely. So let's merge this and do the other issue in a separate PR.

PR #342, opened 2018-03-20
…

Neverhorst:
## Description
Allowed scientific spinbox validator to enter trailing whitespaces after the SI-prefix.

## Motivation and Context
When the LineEdit cursor in the SpinBox  was positioned before a trailing whitespace, one could not enter an SI-prefix. Fixed that.

## How Has This Been Tested?
Dummy

## Types of changes

## Checklist:

### DIFF
--- a/qtwidgets/scientific_spinbox.py
+++ b/qtwidgets/scientific_spinbox.py
@@ -34,7 +34,7 @@ class FloatValidator(QtGui.QValidator):
     Also supports SI unit prefix like 'M', 'n' etc.
     """
 
-    float_re = re.compile(r'((([+-]?\d+)\.?(\d*))?([eE][+-]?\d+)?\s?([YZEPTGMkmµunpfazy]?))')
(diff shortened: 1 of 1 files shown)


Neverhorst:
That is actually not an issue of the widget itself.

The problem here is, that TableWidgets/Views work with an item delegator. Meaning each item in the Table does not have its own widget but the widgets for each item are created on the fly while scrolling and selecting individual items to edit.
The reason for that is to avoid massive resource consumption for large data sets. Imagine a PulseBlock with tens of thousands elements and each parameter of the elements is assigned an individual widget.

Now the tricky part is that the information about the number of digits to show is stored in the widget instance of the scientific SpinBox. So upon losing focus on a table item the underlying widget gets destroyed.
This is an issue I'm trying to fix in the upcoming pulsed overhaul. But I'm not sure yet how... maybe one needs to keep track of the number of digits in the delegate itself.

PR #343, opened 2018-03-21
Pulsed analysis show delta

kay-jahnke:
The GUI for pulsed analysis can show a second plot. Introduced a feature that shows the delta between two measurements in this plot.

Before, the FFT was hardcoded in GUI, logic and measurement. Changed it to have a selection, so it is easier to add more plot types in the future. Also updated the saved pictures to change axes between delta in fft plots.

This feature is only available if an alternating sequence is running.

## How Has This Been Tested?
Tested on dummy: Win7 64x.

## Types of changes

## Checklist:

### DIFF
--- a/gui/pulsed/pulsed_maingui.py
+++ b/gui/pulsed/pulsed_maingui.py
@@ -1714,7 +1714,7 @@ class PulsedMeasurementGui(GUIBase):
         return
 
     def signal_data_updated(self, x_data, y_signal_data, y2_signal_data, y_error_data,
-                            y2_error_data, fft_x_data, fft_y_data, fft_y2_data):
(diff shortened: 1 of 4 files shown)


kay-jahnke:
@Neverhorst I agree with you. I also wasn't to happy with the signal. But if you think connecting through directly is the better way, than let's do it like you proposed in you amendment.

I also was debating if we let the measurement logic always compute all possible second plots and save them. But this might cause a lot of data.

So let's now merge this pull request and wait for your overhaul, to revisit this problem afterwards.

Neverhorst:
Well the proper way in the current state of pulsed would be to add a new signal for that purpose.

But yeah lets revisit.

PR #344, opened 2018-03-21
Added new spectrometer hardware (Andor Camera)

alrik-durand:
I've added a hardware module to interface between the Andor Camera and Qudi via the Spectrometer interface

## Description
This hardware module make use of a python wrapper which is an existing GitHub project (more precisely a fork rewritten in Python 3) : _PyAndor_. 
As submodules in Git can create quite a mess easily and the project haven't been updated in two years, I've included the code directly. (This project is also under GPL v3 licence)
I've created a "libaries" folder for the occasion as it is likely more librairies will be included in the project. It might be a good idea to think about integration guidelines.

I've also included documentation for the hardware module and the interface and included them in the main page.

## Motivation and Context
This change is required because we want to use Andor Camera with Qudi

## How Has This Been Tested?
I've tested this module with current spectrometer logic and spectrometer GUI. 
The camera in a Newton camera via USB on Windows 10 x64. Andor Solis with its drivers is installed in the default folder.
The curve seem to be the same as what Solis (official software) result.
My changes should not affect other modules.

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.


drogenlied:
Please do not add third-party libraries to Qudi, especially if someone already made it installable with setuptools or if someone is hosting a copy on PyPi.
If you want, we can host a version for use with Qudi on PyPi.

Everything else looks quite nice and mergeable and I am happy about the documentation.

alrik-durand:
Ok ! 
I've checked and the "PyAndor" hosted on PyPi and the PyAndor I've used are completely different ! The one from PyPi come from https://github.com/ardiloot/PyAndor whereas the one I've used come from 
https://github.com/hamidohadi/pyandor and they do not seem to be forks of one another.
The other one looked more complete but the licence is not discussed...

I think the better option is that you host a fork of hamidohadi's code on GitHub and PyPi, and then the hardware module code will check if the pip package is installed and the version is up to date (my main worry is that people might update Qudi but not depencencies). This solution is pretty manual but I doubt we will change the libraries code very often.
What do you think about this option ?

drogenlied:
Using and hosting hamidohadi package ssems to be a good idea.

Check `core/util/helpers.py` for the `check_package` function that allows you to check package versions, although you need to move that function outside the function (to the top level) that it is defined in so you can use it.

alrik-durand:
I've forked the repo I was using and I've pushed the updates I've made to  it.
You can fork it and define your version as the reference for PyPi.
`check_package` was exactly what I was looking for !
I'll make the changes to it when the PyPi package is ready.

PR #350, opened 2018-03-26
Pulsed extraction fix

kay-jahnke:
* set the line width of infinite lines to one pixel in order to avoid them disappearing when zooming out.
 * MW-Frequency in the predefined methods tab was hardcoded to 2.87 MHz, changed it to GHz
 * introduced a check for the order of the lines to avoid negative width when moving the wrong line

## How Has This Been Tested?
tested on dummy: Win7 64x

## Types of changes

## Checklist:


kay-jahnke:
Number of commits looks weird (should only have been one), but diff is just as intended.

PR #361, opened 2018-04-06
Save data added numpy filetype

NicolasStaudenmaier:
data can now also be saved as a .npz-file
## Description

The save_data function in the save_logic is extended in order to save (large) data in a compressed .npz-file. The parameters are still saved in a textfile.
## Motivation and Context

much better to save large data
## How Has This Been Tested?

Tested on dummy and setup3
## Types of changes

## Checklist:

### DIFF
--- a/logic/save_logic.py
+++ b/logic/save_logic.py
@@ -408,6 +408,7 @@ class SaveLogic(GenericLogic):
 
         # Reshape data if multiple 1D arrays have been passed to this method.
         # If a 2D array has been passed, reformat the specifier
+        input_data = data
         if len(data) != 1:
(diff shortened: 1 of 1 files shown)


NicolasStaudenmaier:
Thanks for the review. I have moved the reshaping to the if clause for saving as a textfile. I haven't done this before because the reshaped data was used for the header where the identifiers of the data are added. Now in the npz case, the dictionary keys are added to the header instead. This gives a little different format but isn't of any interest because the display of the keys is only for visual reason and not used after all.

NicolasStaudenmaier:
Thanks again, Jan.

PR #362, opened 2018-04-09
fixed a bug where the errors of the delta plot were saved incorrectly

kay-jahnke:
fixed a bug where the errors of the delta plot were not save correctly (or rather the original errors were not saved anymore).

Also included the errorbars into the saved data of the delta plot.

## Types of changes

## Checklist:

- [x] I have documented my changes in the changelog (`documentation/changelog.md`) -> minor bugfixes

### DIFF
--- a/gui/pulsed/pulsed_maingui.py
+++ b/gui/pulsed/pulsed_maingui.py
@@ -1522,10 +1522,6 @@ class PulsedMeasurementGui(GUIBase):
         # apply hardware constraints
         self._analysis_apply_hardware_constraints()
 
-        self.toggle_settings_editor()
-        self.toggle_error_bars()
-        self.change_second_plot()
-
         # initialize values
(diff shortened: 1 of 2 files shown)


drogenlied:
Found two annoying behaviours while testing:

Closing qudi and opening it again with Delta, alternating and errorbrars selected correctly restores the checkboxes but gives the following error:

`Delta can only be selected for the second plot if the sequence is alternating. Setting it to None instead.`
a

Selecting Alternating and then selecting Delta second plot will draw error bars on the delta plot and not save errors.

kay-jahnke:
@drogenlied these bugs are unrelated but worth fixing.

The first one resulted from the order in Initialization in the pulsed gui. I modified the code so the initialization of errorbar and second plot are triggered after everything has been initialized by the measurement file.

The second was just me missing a second clause in the if condition.

Should be all fine now.

PR #366, opened 2018-04-18
Confocal optimizer template fit

kay-jahnke:
A confocal scan can be taken as template for the optimizer and used for optimization by convolution.

## Description
The optimizer setting were extended by one tab. This tab lets you take a template image (XY and Z) and convolute the optimizer with that in order to find the optimal position. The use-case are diamond samples that have fluorescence footprints that deviate from Gaussian (like Nano-pillars, nano-diamond, dense samples).

Also there is a cursor in the template, so you can specify a certain point on the template you want to optimize to. This means, that it is better to take a bit bigger template in order to get more contrasting structure and than choose the correct point on the template.

Along the lines a few minor bugs were fixed and the stability of the optimizer was improved.

## How Has This Been Tested?
New feature has been extensively tested on dummy with Gaussian shaped data.
Win7 64x

## Types of changes

## Checklist:


drogenlied:
Two small things:
* The template is saved, but whether template fitting is used resets on restart
* Optimizer settings shows the template tab by default, not the original optimizer settings tab
* With the dummy config, tracking with POIManager leads to periodic jumping of position by (probably) one pixel.

Someone other than me need to test this with a real sample, and not one of the nice ones.

kay-jahnke:
@drogenlied your first point was intentional to make people look at the template at least, before they fit to it.
Second point was corrected.

Third point I can't reproduce with the confocal only, although it might stem from selecting single pixels. Let's see if this also occurs in real experiments and see if we need to fix it.

kay-jahnke:
I applied a few bug fixes to the z-template-fix (mainly using a different convolution algorithm and fitting the results with a Gaussian).

For testing, I made the dummy data much more noisy.

**The debug messages are still in there, this is intentional for testing. If you give me the go ahead, there will be an additional commit, that kills all the debug stuff.**

a-stark:
Interesting approach, this might be indeed a good way to track nano-pillars, which do not have a gaussian like shape. 
I lack a bit documentation to understand the prodecure. Although it becomes now clear to me what the method `xy_template_fit` and `z_template_fit` is doing, it would be more helpfull to have a short description of the algorithm in the docstring of those methods. I would also add a bit more description to the changelog, since it was not so clear for me what you mean with `[...] and specify offset from that`.
I did not test this on a real setup, just on dummy data.

Neverhorst:
I think this technique has been tested in your labs for quite some time now, @kay-jahnke, is that right?
Could you please add a bit more documentation, as @a-stark requested, so that we can merge this PR after some final checks/testing?

Neverhorst:
__\*bump\*__

kay-jahnke:
Some people have reported that they are not happy with the feature and I don't have the time to fix it without a setup and a proper error description.
So I will kill the PR but leave the branch alive if someone ever wants to use it or if we want to include it into omniscan.

PR #373, opened 2018-04-25
Do not create lots of small ndarrays for POI history

drogenlied:
## Description

The PoI class made a list of ndarrays for coordinates an each of these ndarray was saved to its own file, which is less than ideal.
Now we create a list of lists.

## How Has This Been Tested?

Tested using example dummy config.

## Types of changes

## Checklist:

### DIFF
--- a/logic/poi_manager_logic.py
+++ b/logic/poi_manager_logic.py
@@ -72,7 +72,7 @@ class PoI:
             # Store the time in the history log as seconds since 1970,
             # rather than as a datetime object.
             creation_time_sec = (self._creation_time - datetime.utcfromtimestamp(0)).total_seconds()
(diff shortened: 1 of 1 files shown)


drogenlied:
If performance is still not good enough, we need to put all of it into a single numpy array, which has other problems.

Neverhorst:
That sounds like a similar problem I have with the pulsed objects (`PulseBlock`, `PulseBlockEnsemble`, `PulseSequence`) to be saved as `StatusVar`. 
Lots of nested lists/dicts and the performance is beyond poor, even for as few as 10 PulseBlocks.

But in case of the POIs would saving them as a single numpy array not solve the problem? What is causing trouble in that case?

PR #375, opened 2018-05-03
Bugfix confocal logic gui bridge

latchr:
## Description

Set and Get methods have been added to confocal logic for the scan ranges and scan resolutions. These are now used when confocal GUI gets and sets these scan parameters. Additional signals and update-handling methods have been written, so that now the GUI updates these parameters if they are changed in the logic by something else (console, etc).

Also, the deactivation of the "Stop Scanning" action button in Confocal GUI has been separated from the re-activation of the other scan actions. This makes the GUI state visually map to the real situation, where hardware needs to stop the scan_line before it is truely finished.

## Motivation and Context

The set/get methods for scan parameters is best practice and brings this aspect of the code into compliance with Qudi standards. Also, it is vital for the development of remote GUI solutions.

The scan action buttons state is now better mapped to the hardware/logic state, and this is important for slow scans. Slow scans can arise from slow hardware or from high resolution imaging.

## How Has This Been Tested?

Tested on default config with dummy hardware. These changes only relate to the communication between Confocal logic and GUI modules.

## Types of changes

## Checklist:

### DIFF
--- a/gui/confocal/confocalgui.py
+++ b/gui/confocal/confocalgui.py
@@ -479,12 +479,12 @@ class ConfocalGui(GUIBase):
 
         # Predefine the maximal and minimal image range as the default values
         # for the display of the range:
-        self._mw.x_min_InputWidget.setValue(self._scanning_logic.image_x_range[0])
-        self._mw.x_max_InputWidget.setValue(self._scanning_logic.image_x_range[1])
-        self._mw.y_min_InputWidget.setValue(self._scanning_logic.image_y_range[0])
-        self._mw.y_max_InputWidget.setValue(self._scanning_logic.image_y_range[1])
-        self._mw.z_min_InputWidget.setValue(self._scanning_logic.image_z_range[0])
-        self._mw.z_max_InputWidget.setValue(self._scanning_logic.image_z_range[1])
+        self._mw.x_min_InputWidget.setValue(self._scanning_logic.get_image_axis_range('x')[0])
+        self._mw.x_max_InputWidget.setValue(self._scanning_logic.get_image_axis_range('x')[1])
+        self._mw.y_min_InputWidget.setValue(self._scanning_logic.get_image_axis_range('y')[0])
+        self._mw.y_max_InputWidget.setValue(self._scanning_logic.get_image_axis_range('y')[1])
+        self._mw.z_min_InputWidget.setValue(self._scanning_logic.get_image_axis_range('z')[0])
+        self._mw.z_max_InputWidget.setValue(self._scanning_logic.get_image_axis_range('z')[1])
 
         # Handle slider movements by user:
         self._mw.x_SliderWidget.sliderMoved.connect(self.update_from_slider_x)
@@ -1473,20 +1473,23 @@ class ConfocalGui(GUIBase):
 
     def change_x_image_range(self):
         """ Adjust the image range for x in the logic. """
-        self._scanning_logic.image_x_range = [
+        self._scanning_logic.set_image_axis_range = [
+            'x',
             self._mw.x_min_InputWidget.value(),
             self._mw.x_max_InputWidget.value()]
 
     def change_y_image_range(self):
         """ Adjust the image range for y in the logic.
         """
-        self._scanning_logic.image_y_range = [
+        self._scanning_logic.set_image_axis_range = [
+            'y',
             self._mw.y_min_InputWidget.value(),
             self._mw.y_max_InputWidget.value()]
 
     def change_z_image_range(self):
         """ Adjust the image range for z in the logic. """
-        self._scanning_logic.image_z_range = [
+        self._scanning_logic.set_image_axis_range = [
+            'z',
             self._mw.z_min_InputWidget.value(),
             self._mw.z_max_InputWidget.value()]
 
@@ -1667,10 +1670,8 @@ class ConfocalGui(GUIBase):
         self.refresh_xy_image()
         xy_viewbox = self.xy_image.getViewBox()
 
-        xMin = self._scanning_logic.image_x_range[0]
-        xMax = self._scanning_logic.image_x_range[1]
-        yMin = self._scanning_logic.image_y_range[0]
-        yMax = self._scanning_logic.image_y_range[1]
+        xMin, xMax = self._scanning_logic.get_image_axis_range('x')
+        yMin, yMax = self._scanning_logic.get_image_axis_range('y')
 
         if self.fixed_aspect_ratio_xy:
             # Reset the limit settings so that the method 'setAspectLocked'
@@ -1714,15 +1715,12 @@ class ConfocalGui(GUIBase):
 
         if self._scanning_logic.depth_img_is_xz:
             self._mw.depth_ViewWidget.setLabel('bottom', 'X position', units='m')
-            xMin = self._scanning_logic.image_x_range[0]
-            xMax = self._scanning_logic.image_x_range[1]
+            xMin, xMax = self._scanning_logic.get_image_axis_range('x')
         else:
             self._mw.depth_ViewWidget.setLabel('bottom', 'Y position', units='m')
-            xMin = self._scanning_logic.image_y_range[0]
-            xMax = self._scanning_logic.image_y_range[1]
+            xMin, xMax = self._scanning_logic.get_image_axis_range('y')  # Note 'y'
 
-        zMin = self._scanning_logic.image_z_range[0]
-        zMax = self._scanning_logic.image_z_range[1]
+        zMin, zMax = self._scanning_logic.get_image_axis_range('z')
 
         if self.fixed_aspect_ratio_depth:
             # Reset the limit settings so that the method 'setAspectLocked'
@@ -1753,10 +1751,8 @@ class ConfocalGui(GUIBase):
 
     def put_cursor_in_xy_scan(self):
         """Put the xy crosshair back if it is outside of the visible range. """
-        view_x_min = self._scanning_logic.image_x_range[0]
-        view_x_max = self._scanning_logic.image_x_range[1]
-        view_y_min = self._scanning_logic.image_y_range[0]
-        view_y_max = self._scanning_logic.image_y_range[1]
+        view_x_min, view_x_max = self._scanning_logic.get_image_axis_range('x')
+        view_y_min, view_y_max = self._scanning_logic.get_image_axis_range('y')
 
         x_value = self.roi_xy.pos()[0]
         y_value = self.roi_xy.pos()[1]
@@ -1779,10 +1775,8 @@ class ConfocalGui(GUIBase):
 
     def put_cursor_in_depth_scan(self):
         """Put the depth crosshair back if it is outside of the visible range. """
-        view_x_min = self._scanning_logic.image_x_range[0]
-        view_x_max = self._scanning_logic.image_x_range[1]
-        view_z_min = self._scanning_logic.image_z_range[0]
-        view_z_max = self._scanning_logic.image_z_range[1]
+        view_x_min, view_x_max = self._scanning_logic.get_image_axis_range('x')
+        view_z_min, view_z_max = self._scanning_logic.get_image_axis_range('z')
 
         x_value = self.roi_depth.pos()[0]
         z_value = self.roi_depth.pos()[1]
--- a/gui/confocal/confocalgui.py
+++ b/gui/confocal/confocalgui.py
@@ -520,7 +520,7 @@ class ConfocalGui(GUIBase):
         #################################################################
         # Connect the scan actions to the events if they are clicked. Connect
         # also the adjustment of the displayed windows.
-        self._mw.action_stop_scanning.triggered.connect(self.ready_clicked)
+        self._mw.action_stop_scanning.triggered.connect(self.stop_scanning_clicked)
 
         self._scan_xy_start_proxy = pg.SignalProxy(
             self._mw.action_scan_xy_start.triggered,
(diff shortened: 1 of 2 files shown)


kay-jahnke:
@latchr I like the idea quite a lot, but we need to implement the implementation a bit.

1. The getter/setter methods are outdated by now. Rather use the decorators for getting and setting properties. [Link](https://www.python-course.eu/python3_properties.php)
2. In most things we are now moving to dictionaries, because they have many advantages for us. With you axis parameter (String 'x', 'y', ...) you have nearly done it. I would suggest to just have one getter and setter for example for the range and then using methods to double check the keys of the transmitted dictionaries.
3. I would not call the methods logic_updated, to give us the flexibility to also emit these signals from other points. Jupyter notebooks would be another source, but I can think of more.

If I find time, I might even do a commit or two on this pull request.

I will delay testing on hardware until we have tackled my comments.

latchr:
Good points. Regarding (2) I was thinking in the right direction - you may have noticed the TODO about using dictionaries.

I clearly need to do some reading/learning about decorators for getting/setting. I've quickly browsed the tutorial you linked, and it looks as though it returns the calling syntax to looking like setting a class variable - but with the advantage that signals can be fired in the setter method. 

By the way, I cannot see how "testing on hardware" is required here - only the logic and GUI have been touched.

latchr:
@kay-jahnke I have started converting to dictionaries and I have a question. When I replaced image_x_range with image_range['x'] and so on for all 3 axes, the code would not work. I had to add a declaration of image_range = {} at line 317. Why is this required for the dictionary but it was apparently *not* required for the previous situation?

a-stark:
I am not sure I understand your question correctly, but if you want to include something to a dictionary, you have at first to create that object (=dictionary), which is done via `image_range = {}`. In performing `image_range['x'] = 1234` you actually access the interal setter method of the object `image_range` like `image_range.__setitem__('x', 1234)`.

kay-jahnke:
@latchr Yes your commit is exactly what I meant.

To your question:
You are now using dictionaries, which are essentially named pointers to other "content". So you need to create the dictionary, that hold the pointers first, before you can save the content.
In the earlier version you were storing the content directly in the variable and the "pointing" was done by the naming of the variable implicitly. So in just assigning a value, you took a short-cut and no explicit memory was needed for the actual variable.

As I see it now, you initialize the image_range as an object in the __init__ (line 71) and additionally in the on_activate (line 317) function. Just the __init__ should suffice.

latchr:
@a-stark and @kay-jahnke it is not so simple here because the in it function (line 71) is in the history class, and the on_activate (line 317) is in the confocallogic class. It was not enough to simply declare image_range in the init of the history class. I had to also declare it in the confocallogic class. I am finding it hard to see why it must be declared in both places, but without line 317 it does not work.

Neverhorst:
Sorry if I jump into the discussion here but I get the impression there is something really going the wrong way regarding the idea of future confocal.

The general idea we decided on in several meetings was to generalize confocal in order to perform ANY kind of scanning measurement. This also implies an initially unknown number and labeling of scanning dimensions.
The reason for that is that we now have several reports from lab use cases that highlight how inflexible confocal is when you are not counting with an NI card and move a non-feedback scanner in 3D synced externally to the data acquisition.

The (again) hardcoded channels are going against this idea and so these changes will most probably be removed as soon as the overhaul will take place.
Therefore I would suggest to solely focus on updating the GUI according to logic changes (by introducing a couple of signals) and not change the underlying structure too much in this pull request.
The other alternative I see is to start working on said overhaul but then you need to make a step back and first think of a global confocal structure you want to implement. And this would take much more time.

a-stark:
The instance attribute `image_range` in `ConfocalHistoryEntry` and in `ConfocalLogic` is actually belonging to different objects, but has just the same attribute name in their respective object. 

In [line 71](https://github.com/QMAPP-mq/qudi/blob/939565157d87fd34197b6420300277b03e36e5b9/logic/confocal_logic.py#L71), the created dict object is assigned to an instance variable of the `ConfocalHistoryEntry` object (which is created in [line 323 to 343](https://github.com/QMAPP-mq/qudi/blob/939565157d87fd34197b6420300277b03e36e5b9/logic/confocal_logic.py#L323-L343)). But the instance variable `image_range` in [line 317](https://github.com/QMAPP-mq/qudi/blob/939565157d87fd34197b6420300277b03e36e5b9/logic/confocal_logic.py#L317) is belonging to the `ConfocalLogic` object.

It is actually a bit more complicated. In creating the object `ConfocalHistoryEntry` and passing over the reference from the `ConfocalLogic` (this is done by passing `self` e.g. in [line 323](https://github.com/QMAPP-mq/qudi/blob/939565157d87fd34197b6420300277b03e36e5b9/logic/confocal_logic.py#L323) the attribute `image_range` from the `ConfocalLogic` object is changed by the restore method of the `ConfocalHistoryEntry` object (e.g. in [line 338](https://github.com/QMAPP-mq/qudi/blob/939565157d87fd34197b6420300277b03e36e5b9/logic/confocal_logic.py#L338)). 

Hence if you want to store `image_range` (which is a dict object) in an instance varibale of either `ConfocalHistoryEntry` or `ConfocalLogic` you need to create it first and assign it to the respective object (this was essentially the concise statement of my post above). 

@Neverhorst has a valid point there, that e.g. specifying x y z as image direction could limit in general the applicability of the confocal module. Therefore I would also suggest to keep the current changes of the confocal logic to a minimum (i.e. leaving the image_range as dict and include all desired changes maybe in a dedicated roadmap).

latchr:
I've hit the weekend here in Sydney and won't have time to work on this for a day or two. I'd like clarification on a few issues:

- The @property decorator for getter/setter methods is more pythonic and cleaner. I've learnt this technique playing with this branch today, and it is clearly good. Are we planning to eventually migrate other set/get methods to this style? I hope so.

- @property setters are a bit tricky to think through for dictionaries where only one entry will be edited at a time (such as confocallogic.image_range). Even worse, dictionaries are mutable and must be handled with care when trying to encapsulate with getter/setter methods. I discussed this today at Macquarie with a colleague who has lots of python experience (contributes to the numpy package), and they suggested the current solution with a helper class for ImageRange. This does "hardcode" the axis names x, y, z but this is the current situation in master anyway. How many use cases need non-cartesian axes for confocal logic?

I see this as quite separate from the discussion surrounding the ConfocalScanner interface. That's why I called this branch a bug fix - there were logic parameters which could be changed without the GUI updating to reflect this change. That needs fixing quickly, and the whole thing can be revisited once a new scanning interface is settled.

Neverhorst:
Easiest example for non-cartesian: an angle
But you can think of a lot more when you go away from positional scans.
This has been extensively discussed and will be changed in the future for sure.

Regarding the `ImageRange` class I see another problem. If you use instances of this class in `ConfocalHistoryEntry` the signal `image_ranges_changed_Signal` will emit when you initialize a new history entry or create a snapshot. (In fact they will also emit every time you create a new instance of `ImageRange`)
But this is related to what @a-stark wrote. Image_range in the history entry is different from image_range in the logic.

kay-jahnke:
Hi guy,
I think that spun out of control quite quickly.

In my opinion nobody seems to have time to work on the scanner interface in the moment. Therefore I think smaller and quicker bug fixes might be the way to go, if they improve code and don not need to be done double.

**Concerning the questions:** 
Yes the general idea is to move the code to decorators. But as nobody has time right now, we are implementing the code as we go along.

I liked the way @latchr implemented it earlier by using dictionaries. It would only have taken a few more lines to make it general along the axis.
I also don't like the dummy class ImageRange, because it makes everything even more unflexible and adds another layer of complexity. Rather have one setter function, that takes one or more parameters and than checks their type to either update the existing dictionary keys or even add a new key.

Example:
```
def general_func(**kwargs):
    # kwargs is a dictionary.
    for k,v in kwargs.iteritems():
         print "%s = %s" % (k, v)

def setter_func(**kwargs):
    # check if dictionary was handed over
    if len(kwargs)==1 and isinstance(kwargs[0], dict):
         self.image_range.update(kwargs[0])  # or do some checks and then add the keys from the pasted dictionary
    elif len(kwargs)==2 and isinstance(kwargs[0], str):
         if kwargs[0] in self.image_range.keys():
                  self.image_range[kwargs[0]]=kwargs[1]
         else:
                  print('key not in range')
    else:
         print('cannot handle the parameter of the setter')

general_func(a=1, b=2)
setter_func(['x'=3, 'y'=2])
setter_func('x', 3)
```

Or am I thinking much to complicated?

Neverhorst:
Yep exactly what I was saying earlier. The pull request makes sense but should not complicate the planned overhaul.

By the way... your code example will raise a `TypeError` and `SyntaxError` when you call `setter_func('x', 3)`  and `setter_func(['x'=3, 'y'=2])` respectively.
The last one should probably be a `dict` and not a `list` but in both cases positional arguments are not supported by `setter_func`.

I usually do the same thing you had in mind like:
```
def setter_func(range_dict=None, **kwargs):
    # check if dictionary was handed over
    if range_dict is None:
        range_dict = kwargs
        
    for axis, value in range_dict.items():
        if axis in self.image_range:
            self.image_range[axis] = value
        else:
            print('key not in image_range')
    return
```

kay-jahnke:
@Neverhorst OK, so concerning pull requests, I think we are in agreement.
Also your solution is better, I did not test mine.

I think the majority would also prefer dictionaries as a solution to make the axis arbitrary. So, @latchr if you need help here, please contact us again.

latchr:
@kay-jahnke and @Neverhorst, thanks for the ideas. I have been too busy to touch this for a few days, but I've been thinking about it a lot. The short summary is this: there are some very deep ideas here that impact confocal logic and scanning very broadly! To help move the situation forward, it is important for us to separate two tasks: the simple fix of GUI elements which do not update to reflect logic changes, and a more thorough think through of how all this hangs together.

I am going to close this pull request in order to try and solve these two separate tasks independently. I have a new pull request for a branch called "quickfix...", which implements the required signals to keep the GUI updated to match the logic state for image range and resolution. Based on the helpful feedback in this comment thread here, I have been able to condense the required fixes into the smallest amount of changed lines of code that I think are possible.  In particular, the key idea of using @property decorators to transform established public attributes (image_x_range, xy_resolution, etc) into setter and getter methods has been wonderful. Thanks @kay-jahnke, I've enjoyed learning something new!

This is all that really matters at this stage. But... if you want to understand why the proposed dictionary-based solutions are not optimal without further discussion, then keep reading as I try to outline the issues.

**Dictionaries are not ideal here**
Dictionaries are mutable, which is not what we want for a set of axes in confocal logic. Once the confocal logic is loaded, it is not sensible for a user to change the number of axes (or probably their names) - and yet that is what we would give them by having a dictionary. A simple typo in the console such as confocallogic.axes['x'] = 100 instead of confocallogic.axes['x'].resolution = 100 would overwrite the axis object with the integer 100 - and this is clearly foolish. 

There are some other technical issues here: using a dictionary (designed for collections of a size not known at "compile time" and which can grow) does not allow run-time optimisation tools and other tools such as pylint to do their job properly. This may not be a problem, and dictionaries might make the code easier to read (in other words, they might be worth the drawbacks), but that is not yet clear to me. I would need to see a stronger proof of this.

For example, it seems to me that namedtuple objects are much more suitable for a list of axes that have arbitrary names, but always have the same number of elements (in our case, 3 axes for 3D physical space). I am not experienced with namedtuples, but the discussions I've been having with Python gurus here at Macquarie have convinced me that they are definitely superior to dictionaries in many of the situations where Qudi currently uses dictionaries. We need to be careful to avoid a gleeful adoption of dictionaries everywhere - there are better tools for some jobs.

**Dictionaries do not play nice with @property decorators**
This is a fairly trivial objection, but if you have a look at the thread of comments here you'll see @kay-jahnke saying that "the getter/setter methods are outdated now" and recommending the use of @property decorators. I've read up on this over the last week and definitely agree - the result is more pythonic and cleaner and more elegant to use. However, a few comments further down there are examples of how to implement with dictionaries and they fall back on the "outdated" getter/setter methods. I can understand why, because it is not clear how to write a setter method for just one entry in a dictionary. To see why I would want to do that, imagine me trying to do 

`confocal_logic.image_range['x'] = [0, 10]`

and have a signal emitted whenever this was changed so that the GUI would update its displayed range. 

**It's not just names - confocal logic fundamentally thinks in cartesian coordinates**
There has been a push here to adopt a solution that would allow for arbitrary axes in confocal logic. I still have not been shown any situation where the confocal scanning is done using axes other than "x", "y", "z". While I love the idea of being "as general as we can be", the more I have thought about this the more confused I have become. The confocal scan image data is held in an array that is evenly "sliced up" in two dimensions (ie it is a grid of data). This image data is shown in the GUI as a grid of pixels arranged regularly in a 2D cartesian coordinate system.

If you were to try and do confocal scanning in a polar or spherical coordinate scanner, then it would not be merely the hard-coded "x, y, z" labels that would get in your way. Everything would be messed up. The heart of the confocal logic is the image of data which is built by build_image and then filled by the "scan_line" function. If you try to think about what either of these would be like in non-cartesian coordinates then you will see the problem.  In an r, theta polar coordinate system, the scan_line might scan lines radially out, and the angle theta would vary between subsequent lines. This is fine, but then the current grid of the data image would contain far more pixels at small r and far sparser pixels at large r.

**Perhaps this is trivial to fix - but if so then it is at the hardware module level**
There is an obvious answer to this: "with a polar coordinate scanner we could just do the grid of r, theta values and hold them in the current image that confocal logic has." I agree, although this would mean quite a dramatic visual distortion in the current confocal GUI implementation. The thing is, this solution is already super easy to achieve at one level down from the logic. I could have a spherical coordinate scanner (where can I buy one?) and then simply configure it at the hardware module level to show r as the 'x' axis to the confocal logic, theta as 'y', and phi as 'z' (or any combination I choose). This is actually no different from a physical cartesian scanner showing it's hardware X axis as the 'z' axis in confocal logic, which might easily happen depending on the geometry of the measurement relative to the scanner device.

In other words, if you need to account for scanners with strange axis names or configurations, then this is best handled at the hardware module. At Macquarie we have a real example of this: a scanner with x, y, z axes of sample scan and an additional "tube z" axis that moves the microscope objective along the optical axis. we could choose to do confocal scanning (at the logic level) with the sample x, y, z axes, **or** we could choose to do the confocal scanning with sample x,y axes and the objective "tube_z" axis. In fact, we could choose to do the depth scan (in logic) using the sum of the 'z' and 'tube_z' axes - and increase our depth scan range!  The right place to distinguish and configure between these options is in the hardware module. Once that is done, the hardware module can then "show" 3 axes to the confocal logic and it can happily do cartesian scanning without having to worry about the specifics of the hardware situation. This is the entire point of the hardware abstraction that we really want.

**So finally,...**
It seems to me that if we want the confocal logic to be able to handle arbitrary axis configurations then we will need to drastically rewrite it from the start. I suspect that the best design in this case would be to write a "generic confocal logic" with most of the things, and then inherit this into specific classes for "cartesian confocal logic" and "polar confocal logic" and so on.  This might be fun, it might even be worth it - but it will be a lot of work!

PR #377, opened 2018-05-08
Fix confocal history not setting x scan range spinboxes correctly

### LINKED ISSUE #371: Going back in confocal history messes up x scan range
## *What* is affected by this bug?
x scan range is not properly set by confocal history restore. Sometimes it is set to max range sometimes to zero range (two identical values for start and stop).

## *When* does this occur?
Everytime the confocal history is restored by  `ConfocalLogic.history_back()`.

## *How* do we replicate the issue?
- Wipe app_status
- Start a dummy qudi session with confocal
- set non-maximal scanning ranges for x and y
- start and stop scan to create history entry
- change scanning ranges for x and y
- start and stop scan to create history entry
- Press `Back` button in the toolbar
- x scan range should be different from initial scan settings

## Expected behavior (i.e. solution)
Scan ranges should be updated properly by history events

## Other Comments
I could not find the problem in a quick 15 min search. However it seems not to be related to the `ConfocalHistoryEntry.restore()` method. Calling it manually (via manager console) the proper ranges get set in ConfocalLogic. 
Somewhere along the way during this large signal cascade being emitted by a history restore the values get changed again to the wrong values.

drogenlied:
## Description

Get values for scan ranges from the appropriate place in the logic.
Do not reuse functions that refresh xy or depth image, calling them sets the x range twice.
Also, some style fixes.

## Motivation and Context

Fixes #371.

## How Has This Been Tested?

Dummy only. Please test on actual setup.

## Types of changes

## Checklist:

### DIFF
--- a/gui/confocal/confocalgui.py
+++ b/gui/confocal/confocalgui.py
@@ -36,8 +36,8 @@ from qtpy import QtGui
 from qtpy import QtWidgets
 from qtpy import uic
 
-class CrossROI(pg.ROI):
 
+class CrossROI(pg.ROI):
     """ Create a Region of interest, which is a zoomable rectangular.
 
     @param float pos: optional parameter to set the position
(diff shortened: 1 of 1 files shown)


latchr:
I've had a quick look through this and it seems fine. It certainly has a lot of overlap with #378 , and that makes me wonder which I've been working on for a week. In general, there is no update of the confocal GUI scan range spinboxes when the confocal logic has scan ranges changed. This PR fixes that for scan range, but I don't think it does for the scan resolution?

PR #378 fixes both, but maybe it does not adequately address the problem of Issues #371 ? Can anyone comment further on this after having a look at both pull requests?

drogenlied:
There is some overlap, but this pull request only fixes the history behaviour reported in #371, not the underlying logic/gui communication problem.

Can you try reproducing #731 with #378?

PR #378, opened 2018-05-10
Quickfix confocal logic gui bridge

latchr:
## Description

Set and Get methods have been added to confocal logic for the scan ranges and scan resolutions. These are now used when confocal GUI gets and sets these scan parameters. Additional signals and update-handling methods have been written, so that now the GUI updates these parameters if they are changed in the logic by something else (console, etc).

Also, the deactivation of the "Stop Scanning" action button in Confocal GUI has been separated from the re-activation of the other scan actions. This makes the GUI state visually map to the real situation, where hardware needs to stop the scan_line before it is truely finished.

## Motivation and Context

It is important for the GUI to reflect the state of the logic, and there was a bug where setting the confocal scan ranges and resolutions in the logic was *not* updated in the confocal GUI. This has been fixed in this pull request in the simplest possible way:
  - The GUI relies on the following public attributes of confocal logic: image_x_range, image_y_range, image_z_range, xy_resolution, z_resolution.
  - Set/get methods have been written for these attributes using @property decorators, and so the external facing "API" does not change for the GUI and minimal code changes are required.
  - Two new signals have been added to confocal logic, which are emitted when scan ranges or resolutions (respectively) are changed. The GUI catches these signals and updates its displayed values accordingly.

The scan action buttons state is now better mapped to the hardware/logic state, and this is important for slow scans. Slow scans can arise from slow hardware or from high resolution imaging.

## How Has This Been Tested?

Tested on default config with dummy hardware. These changes only relate to the communication between Confocal logic and GUI modules.

## Types of changes

## Checklist:

### DIFF
--- a/gui/confocal/confocalgui.py
+++ b/gui/confocal/confocalgui.py
@@ -1089,8 +1089,6 @@ class ConfocalGui(GUIBase):
         if self._optimizer_logic.module_state() == 'locked':
             self._optimizer_logic.stop_refocus()
 
-        self.enable_scan_actions()
-
     def xy_scan_clicked(self):
         """ Manages what happens if the xy scan is started. """
         self.disable_scan_actions()
--- a/gui/confocal/confocalgui.py
+++ b/gui/confocal/confocalgui.py
@@ -520,7 +520,7 @@ class ConfocalGui(GUIBase):
         #################################################################
         # Connect the scan actions to the events if they are clicked. Connect
         # also the adjustment of the displayed windows.
-        self._mw.action_stop_scanning.triggered.connect(self.ready_clicked)
+        self._mw.action_stop_scanning.triggered.connect(self.stop_scanning_clicked)
 
         self._scan_xy_start_proxy = pg.SignalProxy(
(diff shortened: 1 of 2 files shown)


latchr:
This is the most condensed "simple/direct solution" to the original problem in PR #375 . The larger ideas growing out of that discussion are best tackled in Issue #379

drogenlied:
In general, this is the way to go right now, except I would split the signals/GUI update functions on a per axis basis.

Right now, restoring a Confocal history state would cause all axes to be set 3 times instead of the required 1.

If you can implement that, I would recreate #377 to take advantage of that update mechanism.

a-stark:
I agree with @drogenlied, splitting the update functionality for each axis is a good idea. I tested your pull request and the behaviour was as expected, but I discovered a bug during testing, which is related to adjusting the image range. Maybe this bug can be handled in this small pull request, if you feel that it fits here: #388
Besides this, it looks good.

Neverhorst:
I will close this PR since there has not been any activity for over a year.
If you drop a PR please also tend to it until it is merged.

PR #380, opened 2018-05-15
New features - camera - steppers - pid - doc

alrik-durand:
## Description
This pull request contain all of the features and changes we started to work on.
Here is what we added to logfile :
* Refined pid interfaces in sub interfaces with inheritance 
* Added a camera interface with logic and basic GUI 
* Added a hardware module for Thorlabs camera 
* Added a hardware module for Andor Newton spectrometer camera for camera, spectro, and setpoint of temperature 
* Added a general multi axis piezo stepper interface 
* Added a hardware module for attocube piezzo steppers 

## Motivation and Context
We began to implement new tools to access an andor spectro, but we've realised that it would be better to create sub modules for the spectrometer, general camera and use PID for temperature control.
The PID interface, logic and gui could be modularised so we did partly in a v_2 folder. 
Detail is given in documentation file
We also used the camera interface to show it can be useful for other camera as a thorlabs camera

We also created a stepper interface and a hardware module for attocube steppers (based on attocube_development branch) but not with scanning in mind (at least not for this interface)

## How Has This Been Tested?
I've tested this changed on a setup we have in Montpellier, France. 
We have an Andor Newton camera for spectrometry, a C1285R12M Thorlabs camera and a ANC 300 with 3 open loop axis.
I've checked that all the reference to modified interfaces have been updated in the code. 
Also, I've check that I can start all module from default config without error.

## Types of changes

## Checklist:


alrik-durand:
Doc string added !

drogenlied:
A few remarks:
* Multiple small pull requests would have made reviewing easier
* Documentation of classes, functions and interfaces should be in the code as docstrings, additional or high level documentation or howtos go in the `documentation` folder in markdown format.
* Documentation is currently compiled into a browseable and searchable format with doxypypy and Doxygen. In may not be easy to build documentation on Windows to check the outcome of documentation changes beforehand. See https://ulm-iqo.github.io/qudi-generated-docs/html-docs/ for the outcome. Things visible on that page need to be added to `main.md`.
* Do you like documentation built with Sphinx better?
* I will post reviews for docs, camera, stepper and PID separately in the next few days

drogenlied:
All of the new interfaces are missing a "dummy" implementation that allows to verify the functionality of logic and GUI modules without hardware.

drogenlied:
logo_qudi.ico is added, but not mentioned anywhere

drogenlied:
Preliminary impression about the different parts:
* Docs need some changes, but most of it is good
* Camera needs a dummy/mock hardware module, but is otherwise probably OK
* Steppers is very specific and probably needs to be renamed to something like "piezo stepper", needs a dummy hardware module
* PID is duplicationg a lot of code. It is not a heavily used module, so direct changes would be fine. However, I do not understand the direction or goal well enough to say if it is a good idea.

Geegee:
I started working on this fork and wanted to thank you for the camera interface, logic and gui.
I expanded the gui  and logic and added hardware file for the camera I'm using (Andor iXon 897 ultra). 
The setpoint interface gives me trouble. To me the idea is to allow hardware which has all kinds
of additional parameters that can or can not be set (e.g. temperature, pressure and so on) which differ 
between hardware devices of the same category (e.g. camera). I like this abstraction. 
The problem right now to me is that you can have only one setpoint variable per hardware
or am I getting this wrong?

I don't know if this is feasible but to me it would be nice if one had a gui in the form of 
a control panel where all these different setpoint values can be set and observed for 
different hardware.

Alrik is there a way to contact you?
Sadly couldn't find a email address here on Github.

alrik-durand:
Hi,
Following the remarks of @a-stark, I've started working on multiple branches to add theses changes more properly. The camera_development and pid_development might be helpful.

I agree that having multiple setpoint has to be a possibility. The way the interface is now is not very flexible, maybe overloading the functions might be a good idea.

Indeed what you are proposing would be nice, but for that we need a clean and general code, not an interface for a specific device (at least I think) this may take some time and effort !

You can email me at alrik.durand@gmail.com

PR #383, opened 2018-05-21
Allow the user to specify a baud rate for the SMIQ GPIB connection

mvbnano:
## Description

- Commit  b2a8e91 allows the user to specify a baud rate for communication with the SMIQ microwave generator. This may be required when using particular GPIB / USB converters.

- Commit 5e8f5c0 adds an example configuration into the class descriptor, to assist new users including the instrument in their config.

## Motivation and Context
If the user is using a GPIB / USB converter that requires a specific baud rate then this value must be parsed into the VISA resource manager to create a connection.

## How Has This Been Tested?
This has been tested creating a connection to the SMIQ06B microwave generator via a [Galvant Industries](http://www.galvant.ca/#!/store/gpibusb) GPIB / USB converter which requires a baud rate of [460800](https://github.com/Galvant/gpibusb-documentation/blob/master/basic_commands.txt).

## Types of changes

## Checklist:

### DIFF
--- a/hardware/microwave/mw_source_smiq.py
+++ b/hardware/microwave/mw_source_smiq.py
@@ -62,7 +62,7 @@ class MicrowaveSmiq(Base, MicrowaveInterface):
         # trying to load the visa connection to the module
         self.rm = visa.ResourceManager()
         try:
-            if not self._gpib_baud_rate:
+            if self._gpib_baud_rate is None:
                 self._gpib_connection = self.rm.open_resource(self._gpib_address,
                                                             timeout=self._gpib_timeout)
             else:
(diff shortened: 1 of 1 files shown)


hardware/microwave/mw_source_smiq.py
```diff
@@ -53,8 +62,13 @@ def on_activate(self):
         # trying to load the visa connection to the module
         self.rm = visa.ResourceManager()
         try:
-            self._gpib_connection = self.rm.open_resource(self._gpib_address,
-                                                          timeout=self._gpib_timeout)
+            if not self._gpib_baud_rate:
```
drogenlied: Please use `is not None` here.

mvbnano: No worries @drogenlied, I have just made that change now in commit 0c48c57.

Issue #300, opened 2018-01-10
Every new extraction or analysis parameter has to be added to the GUI by hand


SimonSchmitt88:
## *What* is affected by this bug?

If there is a new extraction (analysis) method and this method needs parameters which are currently not present in the GUI one has to add a new input box (or similar). Over time this could lead to space problems and limits the overview. 

## Expected behavior (i.e. solution)

So it would be nice if the gui displays only the parameters for the currently chosen extraction (analysis) method (similar of how it is done in the predefined methods or for the magnet)

##Other Comments

Neverhorst:
Fixed by PR #384

Issue #299, opened 2018-01-10
Global parameters in predefined methods tab is not saved when closing pulsedmaingui


SimonSchmitt88:
## *What* is affected by this bug?

When there are changes in the global parameters done they are not saved and when qudi is restarted (the gui is reloaded) the hardcoded values are taken. This is unconvenient and leads to mistakes... Especially after a new sequence should be implemented and during testing multiple reloads might be neccessary.

## *When* does this occur?

After every new start of qudi/ reload of the pulsedmaingui

## Expected behavior (i.e. solution)

When the pulsedmaingui is closed the global parameters should be saved and taken when the gui is reloaded.

##Other Comments

Neverhorst:
Fixed by PR #384

Issue #294, opened 2017-12-01
Make predefined methods more readable/more easy to customize


Neverhorst:
## *What* is affected by this bug?
Each "generate_"-method currently has a lot of additional parameters that are not directly linked to the measurement (like gate_channel, laser_length, etc.). Writing a new generate method includes a lot of code duplication which makes the code less readable. These "infrastructure"-parameters should be somehow handled globally for the sequence_generator_logic.

## *When* does this occur?
When writing custom predefined methods.

## *How* do we replicate the issue?
Just have a look at predefined methods.

## Expected behavior (i.e. solution)
New methods should only have important/physical parameters that are related to the actual measurement sequence.

Neverhorst:
Fixed by PR #384

Issue #239, opened 2017-08-04
Asset name missing in call to _direct_write_sequence


drogenlied:
## *What* is affected by this bug?

For an all-digital pulser device, `direct_write_sequence` cannot be implemented due to asset names missing from the `name` element in the `sequence_params` dictionary.

## *When* does this occur?

Every time a sequence is generated for a device with only digital channels.

## *Where* on the platform does it happen?

See sequence_generator_logic.py, line 944 to 994.

## *How* do we replicate the issue?

- get a dtg3554
- check out dtg3554 branch
- in pulsed gui, create an ensemble
- create a sequence from the ensemble
- try to upload sequence

## Expected behavior (i.e. solution)

The `name` element in the `sequence_params` dictionary should cintain the correct asset names.

Neverhorst:
Fixed by PR #384

PR #384, opened 2018-06-01
Pulsed 3.0

Neverhorst:
## Description
Better late than never...
This huge changeset is motivated by feedback about pulsed measurements over the last year.

Apart from several bug fixes the main focus was on making the pulsed measurement tool chain more flexible and understandable regarding programmatic changes.

Since the most problematic part about pulsed measurements is to ensure consistency of hardware and measurement settings, these changes also aim on providing better access to all kinds of information sets in order to pave the way for more elaborate future measurement implementations.

It also simplifies and streamlines the addition of external pulse analysis and extraction methods as well as analog sampling functions and predefined generate methods.

Responsibility for generating waveforms and sequences in a hardware specific format has been handed over to the pulser hardware modules. This allows for more flexible usage of hardware specific features and removes hardware dependencies from the logic.
These changes required the pulser interface to be altered.

The graphical pulse sequence editors have been revised in order to use the _Qt_ model/view concept.

The "invoke settings" functionality in pulsed measurements should work now properly for pulse sequences generated using predefined methods.

See `changelog.md` for more detailed information.

This is a breaking change. It will require to update the config as well as all custom jupyter notebooks or scripts used to perform pulsed measurements. Old saved `PulseBlock`, `PulseBlockEnsemble` and `PulseSequence` instances are also deleted and can't be restored.

I know that it should be avoided to create such a large pull request but in order to make the modules of the pulsed measurement tool more independent all of them had to be changed at once.
Future changes should be easier to chop into smaller pieces thanks to this pull request.

## Motivation and Context
This change fixes a lot of unreported bugs.

Groundwork for more elaborate pulse analysis and extraction methods.

Better/more stable sequence generation. The actual parameter set for sequences still needs revision but the groundwork is here now.

Fixes issues #300, #299, #294, #239, #93, #87, #86

## How Has This Been Tested?
So far these changes have only been tested on a Win7 x64 dummy configuration.

Especially the changed pulser hardware files will currently not work since these need extensive testing using the actual device. In 2 weeks I will test it on my setup using a Tektronix AWG70002A.
But since this is an unusually big pull request, I wanted to put it up for review of all non-hardware modules as soon as possible.

## Types of changes

## Checklist:


a-stark:
This is a great amount of work @Neverhorst, definitely an improvement! I like very much the Pulse Block Editor, much more compact and more intuitive. Just need. 
So far the dummy configuration works as expected, but I think a more indepth-test with a new hardware device could be the best way to test this new implementation. I suppose that I will need to implement the PulseBlaster and adapt it for qudi. This might be a good test for the new interface/ pulse chain.

drogenlied:
Found a bug with an analog-only instrument I'm implementing right now.

Any channel, digital or analog, can be selected as Laser channel, but sampling an Ensemble fails when an analog channel is selected.

`laser_channel` is checked against the whole `channel_set`, but then `block.element_list[-1].digital_high[laser_channel]` is accessed and fails when `laser_channel` is `a_ch1`.

Neverhorst:
@drogenlied, thank you. 
This should be fixed now (see: 69d30bce465b336fb8f62ce45b8c1569e33c08d9)

Neverhorst:
Hey @a-stark, that's exactly the kind of feedback I was hoping for. I know it's a huge chunk of code for a single pull request so thank you very much for reviewing that.
I will have a detailed look at all of the points you mentioned. 
So far I have the following comments:

1. I had the same thoughts on `write` and `query` and the reason I left it in there was to provide a simple wrapper that returns a more intuitive string (without `\n` or `"`... depending on hardware) and is able to update (error-)status of the device in a qudi conform way or to even block some commands completely.
Of course this might not be needed in the interface. 
So to sum it up: Currently I don't see a reason for these methods to be in the interface but I would like to encourage each hardware module to implement those as (private) helper methods.
The conclusion is then probably to get rid of them in the interface.

2. I can't reproduce the missing unit display in the length and increment columns of the PulseBlock editor. Are you sure they are missing or is the column width simply too small so the suffix gets cut off (it seems so in the screenshot). You can check that by moving the cursor in the spinbox to the right.

3. It is a pain in the ... to properly let Qt handle all sizes of those nested widgets. I fixed the width and height mainly because it's the by far simplest solution. But I see it causes problems...
Another reason for the fixed width/height was the table to behave much more fluent (less resource demanding) when scrolling and populating the table. In my testing this effect was huge.
Do you really need to resize it manually or would it be ok if the table is checking upon initialization what the proper height/width is and fix it then?

4. I started using sets in cases where I often want to compare different sets of strings with each other and avoid duplicates. In those cases the ordering is irrelevant. In some parts of the code these sets need to be cast into lists and there I usually use `sorted()` for type conversion if the ordering is of importance and `list()` otherwise. There is no part as far as I know where it is necessary to maintain the insertion order. So I don't really see a benefit in using `OrderedSet` here. It would even complicate things because you would need to add the elements in the right (alphanumeric) order (which would also need some sort of ordering algorithm in advance).

5. You are right. We need natural sorting here. Maybe the PyPI library `natsort` will help. I will have a look at it. Or maybe it's easier to implement our own channel sorting since we know the structure of the strings (`a_chX` and `d_chX`). That would also avoid another package dependency. I will look into this.

6. The channel config is not really saved as a `StatusVar` but each time you activate the `SequenceGeneratorLogic` it reads the current channel states from the hardware device and updates the currently selected config accordingly. It does not change the hardware settings upon restart. The dummy behaves wrong since it does not save its "settings" as a normal AWG would do. I will adjust the dummy to save the hardware settings as `StatusVar`.
I implemented it in this way because people tend to change hardware settings on the device front panel itself and so qudi will not interfere with that.

7. Will change the doc string. It's `numpy.float32` arrays.
You are right, in most cases I used an unsorted native `dict` (is that even true after moving to newer Python version?) since it does not matter in what order the channels are stored. Also the activation config is a `set` now which also does not care about ordering. Key/value access to these dictionaries does not rely on a specific order in most cases.

8. I see that the description can be a bit more detailed but I'm not sure I understood your point.
What needs to be stated more explicitly? You can (also partially) activate/deactivate channels on your device and the returned state for all channels will always be conform with one of the allowed `activation_config` in the constraints. I will write a bit more in the doc string.

9. That is intended behavior. For pulser devices without their own workspace/mass storage the hardware module will need to provide a "virtual workspace". It can either create a workspace directory on the host PC and simulate AWG behaviour or keep track of the currently uploaded waveform name(s) so it can return them when asked for. 
The first approach has the advantage that you can just sample all waveforms once and store them in a separate directory to be used when clicking `load ensemble`. This would ensure all advantages of a device with integrated workspace. 
The second approach is simpler since you don't need to reserve disk space on your host PC but simply save the waveform name of the currently loaded waveform in RAM. The disadvantage is that you always need to `sample/upload/load` an ensemble when changing the waveform. Simply clicking `load` does not work (unless it's the currently loaded waveform anyway).

10.  Sure, I will include your lines in the comment.
In the currently "normal" use of pulsed the logic will always call `load_waveform` with a list argument. In this case the waveform names need to end on the generic suffix `_chX` in order to invoke channel association from the list of names.
The possibility to pass a dictionary is currently not used by logic modules but is important for future work on the "upload center" we were talking about years ago.
Here you give the appropriate channel number as key and so the waveform can have any arbitrary name (even without the channel suffix).
The list method is more robust since the logic does not need to know how the hardware associates channels to waveforms. Please keep in mind that the term "channel" used in this explanation is not to be confused with what is described as `a_chX` or `d_chX`. In an TekAWG70k for example one waveform is associated with `a_ch1`, `d_ch1` and `d_ch2` at once meaning "channel" 1. This is highly hardware specific.

11. Yeah, I'm also really bothered by the status variable as it is now. (Some other hardware types have the same issue). We should rework that somehow. Is it even needed? Logic modules should actually check return values anyways...
If you have something specific in mind please feel free to propose a change. But I think this can wait until this pull request is through, right?

12. This is a general problem. We still need a lot more sanity checks in pulsed. I added a lot in this change set but we still need many more as you can see in this example. I will check how to improve this specific case but for the future we can simply add more and more checks when we encounter loopholes.

13. Will check/improve that.

Neverhorst:
@a-stark: A short question to point 12.
If I understand correctly then you want the logic to check the value returned by the hardware against the hardware constraints. I think this is not necessary. If the hardware module is implemented correctly it must return the actually set ("physically real") values and they must ALWAYS be within hardware constraints (otherwise you would not be able to set them on the device).
The only case where this could be an issue is when you implement virtual/dummy hardware. So I included range checking in the pulser_dummy module in order to always return sample_rate values within constraints.
Also added the suggested warning for empty waveform generation.
Does that conclude this point then?

a-stark:
to 2: sorry, I think my formulation was a bit misleading. The units are still there, but as you mention correctly, they are hidden by the limited column width.

to 3: I think it would be completely fine, if the tables checks upon initialization the width of the column. 

to 4: ok

to 6: ok, good, just wanted to check that the hardware is the instance, which should save its state.

to 7: from python version 3.6.5 the [dict should have insertion order](https://mail.python.org/pipermail/python-dev/2017-December/151283.html), and this feature will remain now from 3.7 on.

to 8: the point I wanted to mention here is: I assume that the method `set_active_channels(my_active_channel)` in the pulser interface, does not only activate or deactivate the physical channels of a device but in the same run, it should select the activation_config, which is specified for a given channel configuration (here for `my_active_channel`). I.e. it will should also check, whether the given channel configuration is an allowed one, and if not, then the channel config cannot be set. Would you agree with that?

to 9: ok, just to double check: the method `get_waveform_names` should always return a waveform name (at least the currently created one)? Because, if I do the procedure `sample -> upload -> load` (equivalent to the button-press in the gui), I encounter an error in the load method (which calls the `get_waveform_names` to load the current waveform, more precisely in [sequence_generator_logic](https://github.com/Ulm-IQO/qudi/blob/a1403ea7117145ee8d62dcbaa2747e169f7cb0f1/logic/pulsed/sequence_generator_logic.py#L466-L500), with the line [486](https://github.com/Ulm-IQO/qudi/blob/a1403ea7117145ee8d62dcbaa2747e169f7cb0f1/logic/pulsed/sequence_generator_logic.py#L486)) if in my pulseblaster implementation the return value was just an empty list (I copied the behaviour of the [OK pulser]((https://github.com/Ulm-IQO/qudi/blob/a1403ea7117145ee8d62dcbaa2747e169f7cb0f1/hardware/fpga_pulser/ok_fpga_pulser.py)). Therefore I was asking myself, whether the [OK implementation](https://github.com/Ulm-IQO/qudi/blob/a1403ea7117145ee8d62dcbaa2747e169f7cb0f1/hardware/fpga_pulser/ok_fpga_pulser.py#L595-L600) would yield and error? At least the return value (None) would be not correct.

to 10: cool

to 11: yes, let's do the changes concerning the status variable/state in a different pull request, but it is good that we have it on the agenda.

to 12 (and your last comment): hmm I think you are right. It was my bad, that I did not implement the hardware correctly, therefore I was thinking about this little sanity-check-help from the logic (just due to the case that I needed to specify the sample rate manually, as this could not be retrieved from the hardware). But anyway, I would be fine with the warning.


Unfortunately, it was not possible for me to test the pulse extraction and display mechanism of the GUI for fast counting since I currently do not have an actual fast counter at hand. But I assume that you have tested this quite extensively, therefore I think, that this pull request is almost ready to merge.

PR #385, opened 2018-06-07
Improved scientific spinbox validators

Neverhorst:
## Description
Improved the regex of the float validator to be more verbose. It allows for input in a "programmatic" manner, like `.123e12` or `123.e12` now.

A focusIn event (like pressing TAB) on the SpinBoxes will now only select the numeric part of the text excluding the SI-prefix.

Also fixed a subtle bug in the integer validator regex when entering a plus sign in the exponent part.

## Motivation and Context
In some cases the float validator was a bit clumsy. For example one could not delete the last remaining digit of the integer part.

Also when TABing inside the SpinBox the whole number including the SI-prefix was selected. It has proven more convenient if only the numeric part is selected by default since a change of the order of magnitude is less common.

## How Has This Been Tested?
Tested on dummy config (win7 x64)

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -4,6 +4,7 @@
 
 Changes/New features:
 
+* Improved scientific SpinBox validators to allow for more intuitive keyboard input
 * All modules use new connector style where feasible.
 * Bug fix for waveform generation larger than ~2 GSamples
 * Bug fix for POI manager was losing active POI when moving crosshair in confocal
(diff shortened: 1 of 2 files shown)


Neverhorst:
@kay-jahnke, your mentioned "maybe-bug" about entering numbers with leading decimal point while only the numeric text part is selected should also be fixed now.

PR #387, opened 2018-06-15
Correctly generate Y axis for confocal YZ scans

drogenlied:
## Description

Y axis coordinates were not generated from GUI when using YZ depth scan.

Only tested on dummy, needs testing on real setup!

## Types of changes

## Checklist:


kay-jahnke:
While confocal tries to scan YZ, it seems to set the cursor to the XZ position (only during the scan). After the scan the crosshair is set to the correct position.

PR #390, opened 2018-07-12
Desktop shortcut and some new documentation 

alrik-durand:
## Description
I've added a _logo_qudi.ico_ to the _artwork/logo_ folder to help making a nice desktop shortcut on windows with the appropriate explanation in the installation guide.
I've also added some context to the explanation guide to help newcomers understand why install Qudi the proper way.

## Motivation and Context
This change makes having a simple user friendly shortcut desktop on windows easier.

## How Has This Been Tested?
I've used this to make a shortcut and it works great. I haven't change Qudi's main code so I haven't  tested this any other way.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have documented my changes in the changelog (`documentation/changelog.md`)
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have checked that the change does not contain obvious errors (syntax, indentation, mutable default values).
- [ ] I have tested my changes using 'Load all modules' on the default dummy configuration with my changes included.
- [ ] All changed Jupyter notebooks have been stripped of their output cells.

PR #391, opened 2018-07-12
Desktop shortcut and some new documentation  2

alrik-durand:
## Description
I've added a _logo_qudi.ico_ to the _artwork/logo_ folder to help making a nice desktop shortcut on windows with the appropriate explanation in the installation guide.
I've also added some context to the explanation guide to help newcomers understand why install Qudi the proper way.

## Motivation and Context
This change makes having a simple user friendly shortcut desktop on windows easier.

## How Has This Been Tested?
I've used this to make a shortcut and it works great. I haven't change Qudi's main code so I haven't  tested this any other way.

## Types of changes

## Checklist:

### DIFF
--- a/documentation/installation.md
+++ b/documentation/installation.md
@@ -21,7 +21,7 @@ This tool also ease the installation as you can create an environment for Qudi w
 them from the ``.yml`` file in the ``tool`` folder.
 
 It is highly recommended you use this option. If however you want to go mannual, you can find more information 
-[here](@ref required_python_packages)
+[here](@ref manual-package-installation)
 
 #### Git vs Download
 
--- a/documentation/installation.md
+++ b/documentation/installation.md
@@ -74,20 +74,30 @@ https://www.jetbrains.com/pycharm/help/configuring-python-interpreter-for-a-proj
 You can create a Desktop shortcut to launch Qudi easily on your machine.
 
 - On your Desktop, right click and go to ``New->Shortcut``
-- For the location you need to copy the following target :
-    - %windir%\System32\cmd.exe "/K" C:\ProgramData\Miniconda3\Scripts\activate.bat
-    "C:\Users\username\Miniconda3\envs\qudi" &#65120;&#65120; python "start.py"
-        - The &#65120; might cause troube when copy/pasting, you might need to add it by hand
-        - ``C:\ProgramData\Miniconda3\Scripts\activate.bat`` is the Anaconda activation file, it might be in another
-        location depending on your installation
-        - ``C:\Users\username\Miniconda3\envs\qudi`` is the environment folder for Qudi. You can find it by typing
-        ``conda info`` in the terminal while qudi is activated
+- For the location you need to copy the following target 
+    - ` %windir%\System32\cmd.exe "/K" <path-to-activation-script> "<path-to-qudi-environment>"
+     && cd "<path-to-qudi-directory>" && python "start.py" `
+    - In order for the shortcut to work on every windows setup, you need to specify 3 things :
+        - `<path-to-activation-script>` : the path to the Anaconda activate.bat file, for example
+        `C:\ProgramData\Miniconda3\Scripts\activate.bat`.
+        - `<path-to-qudi-environment>` : this can be found using command `conda info --envs` in a terminal.
+        - `<path-to-qudi-directory>` : the path where Qudi's `start.py` can be found.  
 - Click ``Next``
 - Give the name you want fot the shortcut : ``Qudi``
 - Click ``Finish``
-- Right click on the newly created shortcut and go to `Proprerties`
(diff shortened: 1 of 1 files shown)


a-stark:
This pull request is related to #390. Let's continue the discussion here.

Ah, now I get what you wanted to do. The `&#65120` is the unicode character for `&`. In my opinion there is a slighly simpler way. I would use for the shortcut the following syntax:

```
%windir%\System32\cmd.exe "/K" activate qudi && cd "<path-to-qudi-directory>" && python "start.py"
```

The important part here is `%windir%\System32\cmd.exe`, the activate script should be known to the windows path, otherwise it would be not possible (let's say it's cumbersome) to activate the qudi environment. You can also skip the path to the qudi environment, just need the name here. Then you just need to `cd` to the directory of Qudi (e.g.  it may look like <path-to-qudi-directory> = C:\Software\qudi). You do not need to use unicode character `&#65120` here, use them directly as `&`.

The example above works for me (Tested on Windows 7 32bit, 64bit and Windows 10 64bit) and does only require the knowledge of the qudi directory.

Please note, there might be a small pitfall. In Windows you cannot switch directly between partitions with `cd` (i.e. between C: and D:). If the path lays on `D:` then it changes to:

```
%windir%\System32\cmd.exe "/K" activate qudi && D: && cd "<path-to-qudi-directory-on-D-partition>" && python "start.py"
```

....I really do not understand why windows has such a poor terminal

alrik-durand:
I did not know that about windows terminal, I'll add this to the documentation

Ok for "&" !

I have tried what you suggested but it didn't work on my machine. I think when I installed Anaconda I didn't ticked the same cases so "activate qudi" do not work in the windows terminal (conda is known but not the qudi environment). To launch it, we were using the "Anaconda prompt".

Neverhorst:
Just a little experience from my side with anaconda and windows:

The warning in the anaconda installation that ticking the option to add conda to PATH is not recommended has its justification.
I had serious problems on my Win7 x64 machine with Pycharm and conda caused by windows path. That is why I removed conda from path and am only using the anaconda cmd instead.
This method prevents the use of the installation .bat-files (since they use windows cmd) but I had no other problems since. 
Maybe we should consider using Anaconda command prompt by default for install scripts since we require the installation of Anaconda anyways?

a-stark:
I see, @alrik-durand and @Neverhorst, these are valid points!  Indeed, the warning in the installation of Anaconda has its justification (=>"You need to know what you are doing"). The problematic thing with PyCharm and conda concerning the PATH variables in windows is its order, here you can mess up a lot. (It requires a bit manual tweaking that everything works nicely)

But anyway, good that you have mentioned that! To be really unambiguous we need to include the full path to the activation script and the qudi environment, which is then similar to that, what you have proposed @alrik-durand:

```
%windir%\System32\cmd.exe "/K" <path-to-activation-script> "<path-to-qudi-environment>" && cd "<path-to-qudi-directory>" && python "start.py"
```
These would be examples for the paths:
<path-to-activation-script> = C:\ProgramData\Miniconda3\Scripts\activate.bat
<path-to-qudi-environment> = C:\Users\username\Miniconda3\envs\qudi
<path-to-qudi-directory> = C:\Software\qudi

Then we are more independent of the Anaconda cmd. Too bad, I was thinking we could be simpler here.

a-stark:
@alrik-durand, could you check, whether the proposed shortcut is working for you? Maybe you could also update the documentation for that.

alrik-durand:
I don't have access to my computer for a few days, I can do it as soon I get back!

alrik-durand:
Done !
I only have access to a linux machine so I could not test it but if this works on your setup I don't see why it shouldn't on ours.

a-stark:
Thanks @alrik-durand, if there are no other problems, then I would merge this branch at 18:00 (CEST) today.

kay-jahnke:
Hi, for me (win7 64x) it does not work at all:

If I run all three commands (in your description connected by an &&) one after the other it works fine. But if I put them in the link (connected by &&) it gives me the error path not found. 

I also ran the queued command in the command prompt by hand: it did not work with a different error this time. Windows seems to run the three commands at the same time, so it enters the qudi environment but waits for this to finish somehow. If I now exit the qudi environment by typing `exit´ it tries to run qudi but obviously crashes as it is not in the environment anymore.

The only way for me to solve this was to create a batch file that runs the commands and then link to that.

kay-jahnke:
Hm, this way around it seems to be working fine for me.
I do however get the command window with the qudi prompts, but I think this is intended.

a-stark:
Cool, if you have suggestions for improving the description about the shortcut, please let us know.
You will not get rid of the windows command prompt, if you start qudi with it. Another thing is also: If you close the command prompt, you kill qudi. This is the same behaviour if you would close PyCharm (if you use this for qudi).

PR #393, opened 2018-08-14
Save logic threaded saving

### LINKED ISSUE #84: Decouple saving in savelogic via internal signal
Migration: trac#195.

Not feasible right now, we need an interface to specify how we can close plots and how the data to be saved is guarded against changes while the save logic is working on the data in the module wanting to save data.

a-stark:
## Description

Decouple the saving procedure by introducing a queued connection in the save logic. This prevents that Qudi is being blocked during the save procedure. 

## Motivation and Context

Currently, the saving procedure is blocking Qudi if large amount of data is saved (e.g. in confocal).
This PR would fix #84. The Save logic introduces a new signal sigSaveComplete to which other modules can connect if they require to know whether the save is finished. 

## How Has This Been Tested?

Tested with various modules on dummy config. Especially confocal scans will take some time to process the data. By pressing several times the save button you can see that the saved data are processed over time and the experiment can still continue.

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -23,6 +23,7 @@ Changes/New features:
 * Added .ico image to make a desktop shortcut on Windows with explanation in the documentation
 * Added a how-to-participate guide to the documentation
 * Added installation options guide to the documentation
+* Savelogic becomes decoupled from main thread and saving happens in a non blocking way.
 
 Config changes:
 * Add separate conda environments for windows 7 32bit, windows 7 64bit, and windows 10 64bit. 
--- a/logic/save_logic.py
+++ b/logic/save_logic.py
@@ -118,11 +118,17 @@ class FunctionImplementationError(Exception):
 class SaveLogic(GenericLogic):
     """  A general class which saves all kinds of data in a general sense.
 
-    @signal sigSaveData()
+    @signal sigSaveData(object, object, object, object, object, object, object,
+                        object, object, object):
+                        will be used internally to decouple the save procedure.
+    @signal sigSaveComplete(str): Will indicate that the save is complete and
+                                  emit the path where the data has saved to.
     """
 
     sigSaveData = QtCore.Signal(object, object, object, object, object, object,
                                 object, object, object, object)
+
+    sigSaveComplete = QtCore.Signal(str)
 
 
     _modclass = 'savelogic'
@@ -562,6 +568,8 @@ class SaveLogic(GenericLogic):
             plt.close(plotfig)
             self.log.debug('Time needed to save data: {0:.2f}s'.format(time.time()-start_time))
             #----------------------------------------------------------------------------------
+
+        self.sigSaveComplete.emit(filepath)
 
     def save_array_as_text(self, data, filename, filepath='', fmt='%.15e', header='',
                            delimiter='\t', comments='#', append=False):
--- a/logic/save_logic.py
+++ b/logic/save_logic.py
@@ -234,63 +234,60 @@ class SaveLogic(GenericLogic):
         """
         General save routine for data.
 
-        @param dictionary data: Dictionary containing the data to be saved. The keys should be
-                                strings containing the data header/description. The corresponding
-                                items are one or more 1D arrays or one 2D array containing the data
-                                (list or numpy.ndarray). Example:
-
+        @param dict data: Dictionary containing the data to be saved. The keys should be strings
+                          containing the data header/description. The corresponding items are one or
+                          more 1D arrays or one 2D array containing the data (list or numpy.ndarray).
+                          Example:
                                     data = {'Frequency (MHz)': [1,2,4,5,6]}
                                     data = {'Frequency': [1, 2, 4], 'Counts': [234, 894, 743, 423]}
                                     data = {'Frequency (MHz),Counts':[[1,234], [2,894],...[30,504]]}
 
-        @param string filepath: optional, the path to the directory, where the data will be saved.
-                                If the specified path does not exist yet, the saving routine will
-                                try to create it.
-                                If no path is passed (default filepath=None) the saving routine will
-                                create a directory by the name of the calling module inside the
-                                daily data directory.
-                                If no calling module can be inferred and/or the requested path can
-                                not be created the data will be saved in a subfolder of the daily
-                                data directory called UNSPECIFIED
-        @param dictionary parameters: optional, a dictionary with all parameters you want to save in
-                                      the header of the created file.
-        @parem string filename: optional, if you really want to fix your own filename. If passed,
-                                the whole file will have the name
+        @param str filepath: optional, the path to the directory, where the data will be saved. If
+                             the specified path does not exist yet, the saving routine will try to
+                             create it.
+                             If no path is passed (default filepath=None) the saving routine will
+                             create a directory by the name of the calling module inside the daily
+                             data directory.
+                             If no calling module can be inferred and/or the requested path can not
+                             be created the data will be saved in a subfolder of the daily data
+                             directory called UNSPECIFIED.
+        @param dict parameters: optional, a dictionary with all parameters you want to save in the
+                                header of the created file.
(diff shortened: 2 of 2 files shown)


Neverhorst:
Cool, thank you @a-stark.
But the main problem with this method persists, right? If the data to save is changed by the measurement logic while the save logic is processing it we could run into trouble.
Only way to get around that is to ensure that all measurement logic modules wait for `sigSaveComplete` before altering data... and that would lead to the exact same problem as before. Namely that you can not continue with the measurements until the savelogic is finished.
Or did I miss something here? Does the queued connection create a temporary copy of the data?

a-stark:
You are right here. 
The data and parameters are stored in a dict, which is transferred by reference to the save_logic (the reference to the data is maintained through a signal connection, i.e. no copy is created during the signal is emitted). That means eventually if you are running a measurement then the current data at the time of saving are taken, and not the data which were present during the button press event.
And it is true, by introducing a wait until the save has been processed, you will get the same situation as before (without the signal connection).
I believe that each logic should ensure and set the way, how it wants to save data. This can be simply done by literally copy the current data to the data and parameter dict, e.g. by
```
# data_arr is a numpy array containing the current measurement data
data = {}
data['my_data'] = a.copy() # here the current data becomes decoupled.
savelogic.save_data(data)
```
It has to be noted that this will duplicate the data during the save procedure, but at the prize of having the data saved, when the save method of the considered logic module is executed. 

Another approach would be to move the copy procedure of the data before (or directly after) the actual save happens, i.e. in the [save_data method of the save_logic](https://github.com/Ulm-IQO/qudi/blob/88a52eb3373843e89146cfd64aa560df170dca71/logic/save_logic.py#L238-L328). But here, it would be not enough to just copy the passed dict. You would need to iterate through the data in the dict and copy each of them. I am not sure whether you will win here, if you perform the copy of the data at a later stage.

Therefore I would suggest: if it is required that data is saved whenever the save method of a module is executed then it could be solved with a config setting `use_current data`, e.g.:

```
...
if self.use_current data:   # this is a flag, True or False
    data['my_data'] = a.copy()
else:
    data['my_data'] = a
...
```

kay-jahnke:
We have discussed this issues a bit:
Yes, copying the data solves one issue, but for big data and nested dicts this might either blow your memory or be a real pain to do, because you need to reiterate through all the child elements.
Therefore, I would actually say that the current blocking behavior is favorable, as it keeps the user from doing stupid things, without even noticing it.

My suggestion would be to close this pull request without merging and rather work on a way to save the figures (this is what takes so long) thorugh pyqtgraph. I means, they are already rendered and normally we actually don't need very fancy pictures to be saved as default.

latchr:
Perhaps there is an important middle ground solution: I think the GUI locks
up when a slow save process takes place. Even if the logic does need to be
"locked" to prevent stupid things from happening, the GUI should remain
usable.  Is this the way it currently is, or does that need part of this
solution?

a-stark:
Thanks for the input @kay-jahnke, @latchr, but you need to clarify some more things.
I have the impression that the idea of this PR was not understood correctly. 

The proposal of how to handle the saving of data (like mentioned in the examples above) are not part of this PR. These are possible ways to ensure that the data correlates better to the event, when a save action was triggered.
Currently, there could a considerable time difference between when the save event was triggered and when the data are actually written to file.
Therefore, this PR just ensures that the GUI remains responsive.

And for clarification, this has nothing to do with pyqtgraph, unless you really try to capture the displayed graph and save it to file. For all the other parts (confocal, odmr, ...) saving is just done by using numpy and matplotlib.

If you boil down the problematic parts, then there are just two components which take quite some time. One is the saving of the actual data with the numpy method [`np.savetxt`](https://github.com/Ulm-IQO/qudi/blob/9aad6f29443fde93bb8f2f708143f9481cacddc0/logic/save_logic.py#L550-L565), and the other is the saving procedure of the figure within [matplotlib](https://github.com/Ulm-IQO/qudi/blob/9aad6f29443fde93bb8f2f708143f9481cacddc0/logic/save_logic.py#L524).
The choice to execute the native methods was intended and, frankly speaking, I believe that there is very little room for improvement, if you want to save text based with a figure and not with a database (I did not encounter a faster and convenient solution than provided by numpy, please let me know if you have a different suggestion here).


> Therefore, I would actually say that the current blocking behaviour is favourable, as it keeps the user from doing stupid things, without even noticing it.

I disagree with that. On the contrary, the blocking feature is even more dangerous. Say you are doing a big 2D scan with a magnet, you hit the save button and you realize that you need to stop your scan immediately before the stage crashes into the sample. For a scan of 300x300, this can take some seconds (check e.g. a confocal scan), and by this time, your magnet might already hit the sample.

I understand that you want to ensure that no wrong input is done by the user during the saving, but freezing everything is not a solution.


And just as a last comment (which we potentially could put in a separate issue for discussion): If you have a mutable data structure (e.g. numpy array) in where data is continuously altered, then the only way to obtain a snapshot of the data is to create a copy of it. And this is precisely my suggestion. Within each logic, you create a local save method, where a local snapshot of the data is generated, and pass it to the savelogic, which then takes over and saves it. I also cannot see how the user can mess up with this by hitting the savebutton like a machine gun. With a script yes, but this will be even worse in the current blocking version.

Neverhorst:
Just to clarify a few things we talked about in person...

The idea about the current state of the art is that saving is (should) not be possible while a measurement is running (logic checks module_state before saving and GUI disables the save button). Apparently this is not the case for confocal (but it should). See "pulsed" as example.
In short: This discussion is based on the assumption that you can not trigger saving during a running measurement.

Provided saving can not be triggered while a measurement is running (`<logic>.module_state() == 'locked'`), everything is fine. In this case it would be safe to decouple the GUI from the save method call via signal. BUT during the saving process the GUI would need to notify the user somehow that no measurement can be started (which no GUI module currently does). Also the logic module would need safety measures to block a new measurement.
Creating a copy of the data is a valid solution but it is not general. If you have a several GB large (happened already in our institute) array to save this would mess up your resources.

So putting everything on hold while saving is in progress is (currently) the only generally applicable solution.


Regarding the pyqtgraph topic:
@kay-jahnke just had the idea to export the measurement graph directly from the GUI as "measurement thumbnail" instead of using matplotlib because the by far slowest part in the saving process is the figure creation with matplotlib and saving it to disk.
Provided you have a SSD, saving even several GB of data using numpy.savetxt is lightning-fast compared to the figures.

a-stark:
@Neverhorst, thank you for the clarification, now I have the picture of your thinking process. And I can see on which assumption this discussion is based. 
I understand and agree, that copying of data before saving might be not a general way to approach the issue, specifically in the case of several GB of data (yeiks!). The copying process will anyway freeze the logic. But I would most certainly argue that in most of our modules, saving can be implemented like that, and if it is not possible, then within this module it has to be different, and this is all fine (like in the pulsed scenario since big amount of data are expected).

However, there is one thing I disagree with, it is the inability to save at runtime. In my opinion, this is a huge loss in flexibility for the user. I still do not see an issue if e.g. a worker thread is changing a mutable data structure and another saving thread is reading this mutable data structure and saving it (or copying it into another object). Yes, it might be that half of the data are not updated with the most recent data but there should no problems in data corruption or memory access violations. Unless your mutable array is controlled by an external custom c/c++ routine (which needs to take care of proper referencing and dereferencing), you do not need to care about this in python. This is, by the way, one of the advantages of the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock).

With that, I do not see a problem in pressing a save button directly after you have pressed the run button. In either way, you would either get an empty array saved to file (or if you are lucky with the signal processing) you can save the previous data. But this is completely the responsibility of the user.

I do not want to blow up this PR but it might be better to open a separate issue for the discussion on "saving on runtime", to disentangle it from this PR. In this issue, I would be happy to hear about the arguments, why saving at runtime would cause problems, maybe you have already experienced some severe problems, which would support this idea.

PR #396, opened 2018-08-27
I have met with memory stacking up problem, when I tried to save mass…

liuyanulm:
The line added in savelogic, reduces half of stacking up memory from image saving.

The image saving causes memory stack up, more or less on different PCs. This change does not essentially solve this problem, but it reduces half of the memory stack. 
## Description
This line removes current fig in matploitlib.pyplot.

## Motivation and Context
I have met with memory stacking up problem, when I tried to save massive odmr figures. 
computer just goes slower with more and more image-saving, and when memory is fully stacked your measurements would be aborted.
## How Has This Been Tested?
It was tested on my office computer, memory stack reduced from ~20MB/fig to 10 MB/fig.
It was tested in Setup 5, memory stack reduced from ~7 MB/fig to 3.5 MB/fig

## Types of changes

## Checklist:

### DIFF
--- a/logic/save_logic.py
+++ b/logic/save_logic.py
@@ -543,6 +543,7 @@ class SaveLogic(GenericLogic):
             png_image.save(fig_fname_image, "png", pnginfo=png_metadata)
 
             # close matplotlib figure
+            plt.gcf().clear()
             plt.close(plotfig)
             self.log.debug('Time needed to save data: {0:.2f}s'.format(time.time()-start_time))
             #----------------------------------------------------------------------------------


a-stark:
Thank you @liuyanulm, I would suggest the following. Since you already have the reference to the current figure object (`plotfig`) you could directly clear it:
```
plotfig.clear()
```
This would be more deterministic and will not clear by accident another figure (which might be also present in memory). Could you check whether this works for you?

liuyanulm:
plotfig.clear() does the same job as plt.gcf().clear() on setup 5

a-stark:
Another thing just came to my attention: 
What if you want to post process the figure object further in the specific logic, where it was created? If we view the savelogic as the instance which solely saves the figure to disk, shouldn't the parent logic (the logic which creates and passes the figure object to the savelogic) handle also the cleanup? I would have a tendency to believe the latter one, which would mean that the clear operation would move to each parent logic (here odmr logic).
What do you guys think?

liuyanulm:
Actually, I am starting to use some sort of 'solely' saving via jupyter notebook.
see below:
        plt.plot(odmrlogic.odmr_plot_x,odmrlogic.odmr_plot_y[0])
        plt.plot(odmrlogic.odmr_fit_x,odmrlogic.odmr_fit_y)
        plt.ylabel('Intensity (Counts)')
        plt.xlabel('Microwave (Hz)')
        plt.gcf()
        plt.savefig('odmr_{0}.png'.format(*), dpi=150)
        plt.gcf().clear()
For the moment, memory stack does not happen.
By the way, @a-stark , did you also find similar memory stack on your pc?---- Just wonder if it is only happening to me.

Neverhorst:
I had the same thoughts when I revised the SaveLogic a long time ago, just did not went through with it.
The creating logic should also delete the figure in my opinion.

Just another thing. I was not able to reproduce this memory leak issue by looping dummy ODMR measurements and saving them.
However I encountered a lot of strange behaviour with memory allocation when working with matplitlib on windows. All methods to "clear" figures or delete axes etc. do not necessarily free up memory (even when memory runs out), sometimes it just does not free memory at all (no additional references to the figure or anything like that). The problem is that I could not find out when it happens exactly since I could run a script without errors and the next day the exact same script uses 10 times more memory. Excluding matplotlib in these cases always gets rid of this behaviour. I have not found a way to reliably reproduce these issues.

liuyanulm:
images solely saving via jupyter notebook tested overnight on Setup 5, memory did not stack.

liuyanulm:
Hi @a-stark ,
Thank you for the reply.
The memory stacking happens as well when I click saving on GUI with a mouse.
I will try your  suggestion in a next experiment.
The PR merely release half the jammed memory,not really a solution to the issue.
I aggree on your proposal of closing this PR, and will see chances to find a real solution.

PR #408, opened 2018-09-14
Pulsed aftermath bugfixes

Neverhorst:
## Description
Introduced deepcopy operation when adding sequence steps in order to avoid entangled step parameter dictionaries. It's more efficient if the user could take care of copying independent dictionaries before adding them to the PulseSequence object but this way it is bulletproof.

Fixed a bug when sampling a PulseSequence while not preserving rotating frame across  different waveforms. Each different PulseBlockEnsemble should be sampled only once even if they occur in several sequence steps.

Fixed a minor bug in awg70k hardware module related to string formatting.

Added two new status_dict flags in PulsedMasterLogic that could be useful for scripting. One indicates if data fitting is in progress, the other indicates if predefined sequence generation is in progress.

## Motivation and Context
Bugfixes and a minor enhancement

## How Has This Been Tested?
On dummy

## Types of changes

## Checklist:


Neverhorst:
Thanks @SimonSchmitt88, forgot that.

PR #422, opened 2018-09-27
Microwave interface timing

drogenlied:
This change add timing information to the set_ext_trigger function in the microwave interface to enable the use of microwave sources that are level triggered instead of edge triggered and that support a minimum dwell time.

## Description

A timing argument is added to the set_ext_trigger function.
This is supplied by the logic with the time between triggers and can be used to set a dwell time in
microwave sources.

If a device can only be level triggered and supports a minimum dwell time setting, this change makes the device usable for ODMR.

## Motivation and Context

WindFreak Technologies makes microwave synthesizers that can only be level triggered.
A pull request for the WindFreak SynthHDPro device will follow.

## How Has This Been Tested?

Still needs testing on other hardware than WindFreak SynthHDPro.
All other hardware should not have problems.

## Types of changes

## Checklist:

### DIFF
--- a/interface/microwave_interface.py
+++ b/interface/microwave_interface.py
@@ -189,7 +189,8 @@ class MicrowaveInterface(metaclass=InterfaceMetaclass):
         @param TriggerEdge pol: polarisation of the trigger (basically rising edge or falling edge)
         @param timing: estimated time between triggers
 
-        @return object: current trigger polarity [TriggerEdge.RISING, TriggerEdge.FALLING]
+        @return object, float: current trigger polarity [TriggerEdge.RISING, TriggerEdge.FALLING],
+            trigger timing as queried from device
         """
         pass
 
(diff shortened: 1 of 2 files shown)


interface/microwave_interface.py
```diff
@@ -183,10 +183,11 @@ def reset_sweeppos(self):
...
 
     @abc.abstractmethod
-    def set_ext_trigger(self, pol=TriggerEdge.RISING):
+    def set_ext_trigger(self, pol, timing):
         """ Set the external trigger for this device with proper polarization.
 
         @param TriggerEdge pol: polarisation of the trigger (basically rising edge or falling edge)
+        @param timing: estimated time between triggers
 
         @return object: current trigger polarity [TriggerEdge.RISING, TriggerEdge.FALLING]
```
kay-jahnke: please also add the timing to the docu of the return value.

logic/odmr_logic.py
```diff
@@ -208,7 +209,7 @@ def _initialize_odmr_plots(self):
         self.sigOdmrFitUpdated.emit(self.odmr_fit_x, self.odmr_fit_y, {}, current_fit)
         return
 
-    def set_trigger_pol(self, trigger_pol):
+    def set_trigger(self, trigger_pol, frequency):
         """
         Set trigger polarity of external microwave trigger (for list and sweep mode).
```
kay-jahnke: please include frequency in the docu

kay-jahnke:
Thanks, looks good to me, but let someone test it with a setup.

PR #427, opened 2018-09-28
Awg70k constraints

SimonSchmitt88:
Updated the constraints
Make sequence parameters optional
set_digital_level should work with constraints now

## How Has This Been Tested?

AWG70001A on setup 3

## Types of changes

## Checklist:

### DIFF
--- a/hardware/awg/tektronix_awg70k.py
+++ b/hardware/awg/tektronix_awg70k.py
@@ -374,8 +374,12 @@ class AWG70K(Base, PulserInterface):
             self.write('MMEM:OPEN "{0}"'.format(os.path.join(
                 self._ftp_dir, self.ftp_working_dir, wfm_name + '.wfmx')))
             # Wait for everything to complete
-            while int(self.query('*OPC?')) != 1:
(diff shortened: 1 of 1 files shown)


SimonSchmitt88:
Thanks for your review. I removed the try-except clause and increased the timeout duration instead. I tested and the '*opc?'-query is only answered after the loading is completed. So there is no need for a while loop as it has been before, but a single query is sufficient. Since for my setup and very long waveforms, the upload may take close to half an hour the timeout is increased that high...
After the loading is completed, the duration is reset to its previous value.

Neverhorst:
Yes, that should do the trick.

Due to the recent changes in master I resolved some conflicts in the awg70k module and changed the `write_sequence` method accordingly.

Could you please check, @SimonSchmitt88, if this is ok for you? The checks for each sequence parameter have changed in a way that they not check if the parameter is present (since this is ensured by the new class SequenceStep) but if the parameter differs from the device default.

SimonSchmitt88:
Great, thanks. Works for setup 3.

PR #429, opened 2018-09-28
Bugfix and improvement regarding axis labels and units

Neverhorst:
## Description
The handling of axis units and labels in pulsed measurements has been a bit buggy. The "units" entry in the measurement_information dictionary (in predefined methods) has previously been ignored by the GUI.
Now it is not only properly handled but you can additionally set the axis label as well by using this dictionary container with the keyword "labels" (similar to the "units" keyword it requires a pair of strings as values).
Also removed the axis labels and units StatusVars from the GUI. This was redundant with the PulsedMeasurementLogic StatusVars.

## How Has This Been Tested?
Dummy on Win8.1 x64

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -4,7 +4,8 @@
 
 Changes/New features:
 
-* Improved scientific SpinBox validators to allow for more intuitive keyboard input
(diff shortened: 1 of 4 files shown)


kay-jahnke:
improved the implementation to avoid signal loops.

PR #433, opened 2018-10-09
Spinbox unit prefix

alrik-durand:
I've added a feature to the scientific spinbox to have a default prefix.

## Description
The GUI module can set the _assumed_unit_prefix of a spinbox so that a prefix is assumed if none is provided.

## Motivation and Context
For some module, like confocal scan, unit is almost the same. This let the user use the config file to change a spinbox's behavior if this is supported by the GUI module.

## How Has This Been Tested?
I've checked that this does what it's supposed to on a real setup.

## Types of changes

## Checklist:

### DIFF
--- a/config/example/default.cfg
+++ b/config/example/default.cfg
@@ -341,6 +341,7 @@ gui:
         image_x_padding: 0.02
         image_y_padding: 0.02
         image_z_padding: 0.02
+        default_meter_prefix: 'u'
 
     poimanager:
         module.Class: 'poimanager.poimangui.PoiManagerGui'
--- a/gui/confocal/confocalgui.py
+++ b/gui/confocal/confocalgui.py
@@ -485,16 +485,16 @@ class ConfocalGui(GUIBase):
         self._mw.z_max_InputWidget.setValue(self._scanning_logic.image_z_range[1])
 
         if self.default_meter_prefix:
-            self._mw.x_current_InputWidget.fixed_unit_prefix = self.default_meter_prefix
-            self._mw.y_current_InputWidget.fixed_unit_prefix = self.default_meter_prefix
-            self._mw.z_current_InputWidget.fixed_unit_prefix = self.default_meter_prefix
+            self._mw.x_current_InputWidget.assumed_unit_prefix = self.default_meter_prefix
+            self._mw.y_current_InputWidget.assumed_unit_prefix = self.default_meter_prefix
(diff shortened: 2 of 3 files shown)


Neverhorst:
I see potential benefit of this PR, however I'm not quite comfortable with the user not being able to enter values without SI-prefix if you set the assumed prefix property.

The discussion should be about if you want to have an SI-prefix when your `value == 0` (this would help operating confocal) or if you want to be able to set a fixed prefix without the ScienDSpinBox rescaling it. (I think this would be the less ambiguous solution and your PR is almost at this point).

If you just want to avoid selection of the entire text (including prefix) inside the SpinBox, you can already get around that by using TAB to select the SpinBox. This will only select the number part of your string without prefix. The only case currently where using TAB will not avoid entering the prefix by hand is when you have zero as value (see previous paragraph).

Long story short: Assuming you use TAB to select the SpinBoxes would it not solve all problems if the zero would maintain the SI prefix?

alrik-durand:
I see your point and indeed this can be a problem in some cases.

This was requested by multiple persons, and I've used this feature for some time and in my case it was quite usefull.

The zero trick could work but I'm worried that when using the slider, the last prefix saved would be nano (in my case) instead of µ.

Neverhorst:
Well I'm not really against it, just thinking if we could make this without excluding a certain value range.

Do you think the other point I mentioned would work? Being able to fix the SI prefix for a ScienDSpinBox?

alrik-durand:
I'm not sure if I understand well what you're proposing. 
The concern I have is : If you set the position to zero by dragging the cursors in the *scan parameter*, and then type a value without prefix, is it going to guess right ?

Neverhorst:
I just checked and you are right. If you drag the slider towards zero, the last number shown has a nano-prefix (or even smaller for longer sliders). So that's not an option.

Another possibility just came to my mind:
What if we just alter your current solution slightly so that it is not checking for a missing unit prefix but instead it checks whether the entered number (as it is, so for example 100) lies within the allowed min-max range. If the number is too small/big it tries to add the default prefix and checks again. If it's still outside the boundaries, it behaves as usual and sets the allowed min or max value instead.
This behavior is more flexible to my mind and does not exclude a certain parameter range. In your particular use case it should also not make a difference.

alrik-durand:
I've implemented your idea
This works on my dummy config

PR #440, opened 2018-10-29
bug fix in double and triple sine

SimonSchmitt88:
When thinking of a Double or Triple - sine, I would expect the sum of three or two sinuses and not the product.

## Description

Changed the product to a sum

## Motivation and Context

When a signal with two or three different frequencies should be applied the result is the sum of the three individual signals 

## How Has This Been Tested?

Tested on setup 3.

## Types of changes

## Checklist:

### DIFF
--- a/logic/pulsed/pulse_objects.py
+++ b/logic/pulsed/pulse_objects.py
@@ -1223,7 +1223,7 @@ class PredefinedGeneratorBase:
                     frequency=freqs[0],
                     phase=phases[0])
             elif sine_number == 2:
-                mw_element.pulse_function[self.microwave_channel] = SamplingFunctions.DoubleSin(
+                mw_element.pulse_function[self.microwave_channel] = SamplingFunctions.DoubleSinSum(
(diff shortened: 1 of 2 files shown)


Neverhorst:
It was a sum of sines before (like a year ago or so) and then people (I think it was @kay-jahnke, sorry if not) urged me to change it to a product.
Please settle on one solution and define a use case for that before we change it back in another year...

SimonSchmitt88:
Thanks for your comments. For me it seems most user-friendly, that a signal, generated by entering three different frequencies, also consists of these frequencies (by means of FT).

An example, when a linear superposition is needed, is if two hyperfine transitions (freq1 and freq2) should be driven simultanously. 

Anyway, I added the product version again and changed the function names accordingly. In the other part of the main branch, I only found one appearence of these functions ('_get_multiple_mw_element' in 'pulse_objects.py') and here I changed to the sum version (for the reasons described above).

I hope this is fine for everyone. If not please let me know.

PR #442, opened 2018-11-08
Bulse blaster delays

alrik-durand:
## Description
I've added a feature for one hardware module : spincore/pulse_blaster_esrpro to take account the line delays at this level. 

## Motivation and Context
When controling an AOM or a MW switch with the pulse blaster, delays may need to be taken into account for precise operation. Letting Qudi take care of this helps us forget all about this problem.
I would have liked to implement this on the logic level, but the pulsed modules have lots of options that makes it complicated.

## How Has This Been Tested?
I have tested this on a real setup with a PBESR-PRO-400 (400 MHz) with Windows 7.
If the new config field is left empty, no breaking change should occur.

## Types of changes

## Checklist:

### DIFF
--- a/hardware/spincore/pulse_blaster_esrpro.py
+++ b/hardware/spincore/pulse_blaster_esrpro.py
@@ -93,11 +93,11 @@ class PulseBlasterESRPRO(Base, SwitchInterface, PulserInterface):
         _channel_delays: Specify the delay of some channels to correct it automatically
                          by this module. For example :
                           channel_delays:
-                                '2': 200-9
-                                '3': 500-9
-                         tell the card the line 2 and 3 have a delay of 200 ns and 500 ns respectively
+                                '0': 200-9
+                                '2': 500-9
+                         tell the card the line 0 and 2 have a delay of 200 ns and 500 ns respectively
                          so that the pulse are emitted sooner relatively to other channels.
-                         The first line is 1 and the last is 21.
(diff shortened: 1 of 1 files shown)


a-stark:
Thanks @alrik-durand for this input. I am not against this option to have a configurable delay for the operations. However, I see this more suitable in the logic part (as @Neverhorst mentioned).

Before merge, I would suggest some small corrections to the doc strings:
- the [underscore in the description](https://github.com/Ulm-IQO/qudi/blob/05bb66369934fb8992d8802caf23a681da90c395/hardware/spincore/pulse_blaster_esrpro.py#L93) of the config option should disappear.
- Please add to the doc string in the method [_correct_sequence_for_delays](https://github.com/Ulm-IQO/qudi/blob/05bb66369934fb8992d8802caf23a681da90c395/hardware/spincore/pulse_blaster_esrpro.py#L932) the parameter description (and return parameter). Or at least reference it to a method, where this parameter is described. If I would read this function as a new user, I would need to understand the logic of the function to get the correct parameters, and this is time consuming and inefficent. Potentially you could just [copy paste this doc string](https://github.com/Ulm-IQO/qudi/blob/05bb66369934fb8992d8802caf23a681da90c395/hardware/spincore/pulse_blaster_esrpro.py#L685-L694).

alrik-durand:
Thanks for your comments,
I've modified the doc strings.

Issue #58, opened 2017-02-02
There is no documentation to show what is available for inclusion in the config file


drogenlied:
Migration: trac#44.

There is no easy way to know what is available for inclusion in the config file. Up until now the default.cfg has been a sort of "catch-all" exhaustive list, but this makes less and less sense as we get more real hardware modules.

drogenlied:
This is not quite true. In the tools directory there is a listmods.py tool which shows the available modules for including in config. However, this process should still be documented in a way that makes it easier for an "outsider" to step through setting up QuDi? on their own experiment.

liuyanulm:
Documentation has been implemented. closing this issue.

Neverhorst:
> This is not quite true. In the tools directory there is a listmods.py tool which shows the available modules for including in config. However, this process should still be documented in a way that makes it easier for an "outsider" to step through setting up QuDi? on their own experiment.

There is still no documentation about where to find all available modules and how to include them in the config.

PR #443, opened 2018-11-19
Copy-paste config example in the docstring of the hardware modules

a-stark:
## Description

Add a config example to each hardware module which can be copy-pasted.

## Motivation and Context

Currently, the config options do not specify a data type, hence, it is very hard to understand the config option if no examples or descriptions are present. (I believe that an attempt in this direction was pursued by @latchr.)
For qudi users, it might be just tedious to find out the correct type of the config option, because the simple 'shape' (datatype) is not always obvious and it forces you to go through the implemented methods, to extract this information (=> time consuming).
For new qudi users, it increases significantly the barrier to start with an implemented module.

Therefore, I propose to insert (at least in every hardware) qudi module an example configuration, which can be just copy-pasted into the config file. The biggest advantage is that the config pattern is already given, and the present options have just to be adapted.

This pull request contains mainly docstring additions, therefore the changes should be non-breaking. Only the [AWG5002C](https://github.com/Ulm-IQO/qudi/blob/0a8868e502b93965b9bc606b29c1d0425e61bdfd/hardware/awg/tektronix_awg5002c.py) file was adapted to have the same pattern for the pulsed_file_dir option as in [tektronix_awg70k.py](https://github.com/Ulm-IQO/qudi/compare/configexample_in_hardware?expand=1#diff-85a9070b02c419b3e73d1f4fc89f0680) and [tektronix_awg7k.py](https://github.com/Ulm-IQO/qudi/compare/configexample_in_hardware?expand=1#diff-72e8335b98fce72631bc45b20a3421bf).

## Types of changes

- [x] Improve docstrings (non-breaking changes)

## Checklist:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -27,6 +27,7 @@
 - [ ] I have documented my changes in the changelog (`documentation/changelog.md`)
 - [ ] My change requires a change to the documentation.
 - [ ] I have updated the documentation accordingly.
+- [ ] I have added/updated for the module the config example in the docstring of the class accordingly.
 - [ ] I have checked that the change does not contain obvious errors (syntax, indentation, mutable default values).
 - [ ] I have tested my changes using 'Load all modules' on the default dummy configuration with my changes included.
 - [ ] All changed Jupyter notebooks have been stripped of their output cells.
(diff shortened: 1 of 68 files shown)


Neverhorst:
Can we say this PR resolves issue #58?

drogenlied:
@Neverhorst  only partially, this one is only for hardware.

For logic and GUI, this problem is still open.

I'm going to merge this before someone adds more hardware without his sort of documentation.

a-stark:
Jap, I agree with @drogenlied, this solves only the problem for the hardware part. But still, it is a step in the right direction. I would still close this PR since it is very broad and its solution might include several steps (including a config editor, which is still work in progress). I would be in favour of opening a more specifc issue on this.

PR #446, opened 2018-11-22
Live odmr

kay-jahnke:
In the ODMR you can now select of how many lines the average should be taken.

## Description

This feature is quite useful if you want to do magnetic field alignment on a live running ODMR. If you want to still average over all the line (this is the default) just put the number of lines to 0.
You can also change the number of line, when ODMR is not running.
The data saved will also be the one displayed in the GUI.

## How Has This Been Tested?

Tested on Setup Quantentunnel with Win10 1803. Tested on Lab Polarizer Win8.1

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -24,6 +24,7 @@ This can be used to specify the axis labels for the measurement (excluding units
 * Microwave interface passes trigger timing to microwave source, needs hardware module adjustments for not-in-tree modules
 * Bug fixes and support for SMD12 laser controller
 * For SMIQs added config options to additionally limit frequency and power. Added constraint for SMQ06B model.
+* Added live OMDR functionality to only calculate the average signal over a limited amount of scanned lines
 * New hardware file for Microwave source - Anritsu MG3691C has been added.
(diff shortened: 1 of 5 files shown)


kay-jahnke:
@Neverhorst Thank you very much for your comments.

They are good points and I agree with all of them.

Two comments:
- I do not see a usecase, for the lines_to_average to be changed from logic apart from the StatusVar, so I now only implemented the initialization. We can however also implement, that the gui is changed for logic changes, but currently I don't see the point.
- I placed the GUI element where it fit best, but your right, logically it lives somewhere else, if just makes the window bigger.

More comments?

Neverhorst:
Thank your @kay-jahnke.
It has been decided some time ago that all logic elements that have a corresponding GUI representation need to update the GUI upon a value change.
It's easily done since there is already a signal for that in the logic (`sigParameterUpdated`).
I took the freedom to change that (see commit bd7897f).

The GUI looks good now but could you please revert the first horizontal line to the full width of the dockwidget?

kay-jahnke:
Thanks, fixed the line. However that happened.

PR #447, opened 2018-11-22
Improvement for loading assets into pulser

Neverhorst:
## Description
Clearing the pulse generator or loading a new asset while the output is still active will now stop the pulser and issue a warning.

## Motivation and Context
It has been inconvenient that you were not able to load a new asset into a running pulse generator without an error. Now you will only receive a warning and the pulser will be stopped.

## How Has This Been Tested?
On dummy Win 8.1 x64

## Types of changes

## Checklist:

### DIFF
--- a/logic/pulsed/pulsed_master_logic.py
+++ b/logic/pulsed/pulsed_master_logic.py
@@ -647,6 +647,10 @@ class PulsedMasterLogic(GenericLogic):
         if still_busy:
             self.log.error('Can not clear pulse generator. Sampling/Loading still in progress.')
         else:
+            if self.status_dict['pulser_running']:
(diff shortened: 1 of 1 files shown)


kay-jahnke:
Tested it on Quantentunnel and it works as intended.

I do have one remark however: If you have an experiment running (not only just the pulser) it turns off the pulser, but leases the experiment itself running. I think the experiment should also be stopped in this case.

Neverhorst:
You are right, @kay-jahnke, I did not think of this case.
However I changed it in a way that it will not load/clear but throw an error when a measurement is running. I think this is preferred behaviour since the measurement should always have highest priority and be terminated explicitly. What do you think?

PR #449, opened 2018-11-22
Pulsed delta analysis

Neverhorst:
## Description
Added the functionality to use the _show Delta_ CheckBox in the _analysis_ tab of the pulsed maingui.
You can now switch between analyzing an alternating measurement as two separate data sets or as a single differential data set.

## Motivation and Context
Fitting and other things could only be applied to the main data plot in pulsed measurements.

## How Has This Been Tested?
On dummy Win8.1 x64

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -116,6 +116,8 @@ This can be used to specify the axis labels for the measurement (excluding units
 	  the ungated trace. For fine-tuning additional delays (for example from AOMs) can be taken 
 	  into account. This method speeds up laser extractions from ungated timetraced by a lot.
 	* Improved pulsed measurement textfile and plot layout for saved data
+	* When an alternating measurement is running the CheckBox "show Delta" can be used to analyze 
+	  the absolute difference of the alternating data sets
     
 
 Config changes:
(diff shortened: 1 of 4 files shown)


kay-jahnke:
The functionality works in principle and I agree that you would either fit the delta or the single curve (actually currently only one of them).

The UI interaction is weird however: we have the second plot, where either the delta or the FFT is shown. The checkbox "Show Delta" has nothing to do with the second plot but rather manipulates whats shown (and fitted) in the first plot.

My suggestion is to rather make a checkbox that switches the fit between both plots. This way you could even fit the FFT if you wanted.

@Neverhorst What do you think?

Neverhorst:
The idea was that alternating measurements are some sort of special case that can be analyzed in two ways, either as two data sets or as difference (thus the CheckBox). 
I would even suggest getting rid of the "delta" in the alternative plot... you just got used to it being a secondary plot option. 
The main plot was intended to be used for the measurement analysis and the second plot was just "nice to have" for visual support.
If you want to have full analysis capability on both plots... why not introduce a third (alternating, delta & FFT) or four (alternating, alternating FFT, delta, delta FFT)?
I think that gets out of hand and really specific advanced and customized data analysis should be done offline.

Neverhorst:
In addition you would need fit result windows for each plot

Neverhorst:
Ok, since fitting on both plots requires an entirely different approach than this PR I will open up a new branch and PR soon.

PR #450, opened 2018-11-23
Interfuse to control AOM with analog voltage

alrik-durand:
## Description
I've added an interfuse to control the power of the laser beam with the laser logic, with an AOM driver.

## Motivation and Context
Some acousto optic modulator (AOM) can be controled with an analog voltage as a second input
to control the output power of the laser beam. This interfuse use the forth analog ouput of the confocal scanner interface ('a' of 'xyza') to send a control voltage on the AOM.

## How Has This Been Tested?
This was checked on one of our setup.

## Types of changes

## Checklist:

### DIFF
--- a/gui/laser/laser.py
+++ b/gui/laser/laser.py
@@ -243,7 +243,7 @@ class LaserGUI(GUIBase):
         elif self._laser_logic.laser_shutter == ShutterState.NOSHUTTER:
             self._mw.shutterButton.setText('No shutter.')
         else:
-            self._mw.laserButton.setText('Shutter: ?')
+            self._mw.shutterButton.setText('Shutter: ?')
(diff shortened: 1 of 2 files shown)


kay-jahnke:
I like the idea, although I don't have an AOM-Driver that could test it.

Two remarks however: You are calling `self._scanner.scanner_set_position(a=voltage)` (line 246) without any checks. You should definitely check before if the scanner is set up correctly and not used by another tool.

Currently you return LaserState.UNKWOWN. Why not set up an internal state variable and then set the voltage to 0 for state "OFF" and to whatever the power was set to before for state "ON"?

alrik-durand:
Everything is working well now !

Issue #441, opened 2018-11-06
Buttons to delete all saved blocks, ensembles and sequences


SimonSchmitt88:
## *What* is affected by this bug?

If a lot of blocks, ensembles or sequences have been saved, it is annoying and time-consuming to find a specific one in the GUI. For deleting individual blocks, ensembles or sequences , there is a button, but deleting more than 100 blocks, ensembles or sequences this way is also annoying. Of course a for-loop (or similar) can be used but it would be nice to have a button 'Delete all' for saved blocks, ensembles and sequences.

PR #451, opened 2018-11-23
Delete all pulsed assets buttons

Neverhorst:
## Description
Introduced new buttons in each pulse editor to delete all saved assets (PulseBlock, PulseBlockEnsemble, PulseSequence) and not just a single one.
The user will be asked to confirm the deletion.

## Motivation and Context
It can be tedious to delete a large number of saved assets one by one. The other way to delete efficiently a large number is to delete the corresponding serialized files in the "saved_pulsed_assets" folder while the SequenceGeneratorLogic is deactivated.

Fixes issue #441 

## How Has This Been Tested?
On dummy Win8.1 x64

## Screenshots:
blockeditor
sequenceeditor

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -117,7 +117,7 @@ This can be used to specify the axis labels for the measurement (excluding units
 	  the ungated trace. For fine-tuning additional delays (for example from AOMs) can be taken 
 	  into account. This method speeds up laser extractions from ungated timetraced by a lot.
 	* Improved pulsed measurement textfile and plot layout for saved data
-    
+    * Added buttons to delete all saved PulseBlock/PulseBlockEnsemble/PulseSequence objects at once.
 
 Config changes:
 * **All** pulsed related logic module paths need to be changed because they have been moved in the logic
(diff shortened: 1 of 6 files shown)


Neverhorst:
Thank you @drogenlied. Failing to deserialize an ensemble/block/sequence will now delete the corresponding file. This is independent of the "delete all" functionality. (I can not just delete everything that is in the folder)

@kay-jahnke, yes, deleting a PulseBlockEnsemble or PulseSequence should also delete the corresponding waveform/sequence on the pulser. This should avoid unpredictable mismatches of abstract pulsed object and waveform/sequence. If the corresponding waveform/sequence is currently loaded in the pulse generator this will result in stopping the device.
I would like to keep it that way but I could add this to the warning or prevent the loaded asset from deletion if it is loaded (but in that case I would also keep the PulseBlockEnsemble/PulseSequence).

Neverhorst:
Try now, @kay-jahnke.

Neverhorst:
@drogenlied: Does that address all issues with this PR?

PR #455, opened 2018-11-28
Pulsed alternative plot fit

Neverhorst:
## Description
Introduced separate fitting widgets to each plot in pulsed measurements. (main plot and alternative plot)

## Motivation and Context
There has been a request for fitting capability on both plots

## How Has This Been Tested?
On dummy Win8.1 x64

capture

## Types of changes

## Checklist:


kay-jahnke:
@Neverhorst Thank you very much.

Tested on dummy win7 64x.

If I want to save the delta plot I get the following message and no fit is saved in the second plot (saving works however). For FFT it works for me:
`No handles with labels found to put in legend.`

Also if I change what is plotted in the second plot, the old fit line stays where it is and therefore avoids rescaling. Can't we just get rid of the fit, once we change what is plotted in the second plot (maybe by calling Fit with the function No Fit.

Additionally I noticed, that what you see is what you get: if no fit or an old fit was performed on the second plot it is also saved this way. I think this is definitely the way to go, as anything else will lead to inconsistencies.

PR #458, opened 2018-11-29
Bugfix: ensemble length in bins and waveform length mismatch

Neverhorst:
## Description
Fixed a very subtle bug that caused the length of the `PulseBlockEnsemble` shown in the editor in bins to differ from the actually sampled waveform length by +-1 bin.

## Motivation and Context
This mismatch has not been reported so far and occurred now with a certain very long waveform.
The mismatch was in the output of the `SequenceGeneratorLogic` methods `analyze_block_ensemble` and `get_ensemble_info`. Both return the `PulseBlockEnsemble` length in bins but in rare cases these numbers differ due to a subtle difference in the calculation method.
`analyze_block_ensemble` is a much more complex computation that was only used for wavfeform creation (sampling) and that's why `get_ensemble_info` has been used to display the information in the GUI.
Now everything uses the more complex method but `get_ensemble_info` is still there in case someone uses it in a script but is effectively just calling `analyze_block_ensemble`.

## How Has This Been Tested?
Cryo lab setup and dummy Win8.1 x64

## Types of changes

## Checklist:

### DIFF
--- a/logic/pulsed/sequence_generator_logic.py
+++ b/logic/pulsed/sequence_generator_logic.py
@@ -1036,66 +1036,29 @@ class SequenceGeneratorLogic(GenericLogic):
     # ---------------------------------------------------------------------------
     def get_ensemble_info(self, ensemble):
         """
-        This helper method will analyze a PulseBlockEnsemble and return information like length in
-        seconds and bins (with currently set sampling rate), number of laser pulses (with currently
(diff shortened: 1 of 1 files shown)


Neverhorst:
That's no problem @kay-jahnke. I tested it with the waveform we encountered the bug with.

PR #459, opened 2018-12-04
Lock-In ODMR

kay-jahnke:
Added the capability to oversample the counter on the ODMR for the NI-Card. Thereby, a switch can be added after the microwave source for fast microwave switching which then let's you measure the delta between MW ON and MW OFF.
Also made the Anritsu microwave source hardware file more general to cope with more microwave sources.

## Description

Added two parameters to ODMR: first to determine the oversampling and the second to activate the lock in itself. There will be a warning that tell you which outputs to connect the triggers to.
The signal with Lock-In active is the ODMR contrast calculated from the differential signal and normalized to the counts at microwave off.

## Motivation and Context

In some experiments the detected signal varies quite a lot. This is especially true for detecting analogue signals in the ODMR. Here it is more beneficial to additionally switch the microwave off and on very fast and then compare the signals from MW OFF and ON. This gets rid of all noise slower than the frequency of oversampling. If oversampling is done to fast at low laser powers the oversampling is not to be done to fast, because otherwise the NV has not enough time to be polarized, reducing the contrast.

## How Has This Been Tested?

Tested in Qunatentunnel Win10 64x 1803 and digital counting.
Testing on LP-Optics Win10 64x 1803 and analogue counting.

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -26,7 +26,7 @@ This can be used to specify the axis labels for the measurement (excluding units
 * For SMIQs added config options to additionally limit frequency and power. Added constraint for SMQ06B model.
 * Added live OMDR functionality to only calculate the average signal over a limited amount of scanned lines
 * New hardware file for Microwave source - Anritsu MG3691C with SCPI commands has been added.
-* **Potential Config Change:** Hardware file for mw_source_anritsu70GHz.py with class MicrowaveAnritsu70GHz was changed to file mw_source_anritsu_MG369x.py with class MicrowaveAnritsuMG369x to make it universal. Also hardware constraints are set per model.
+* **Config Change:** Hardware file for mw_source_anritsu70GHz.py with class MicrowaveAnritsu70GHz was changed to file mw_source_anritsu_MG369x.py with class MicrowaveAnritsuMG369x to make it universal. Also hardware constraints are set per model.
 * Lock-In functionality was added to the ODMR counter and implemented for the NI-Card. All other hardware and interfuse with ODMRCounterInterface were updated.
 * New hardware file for Microwave source - WindFreak Technologies SynthHDPro 54MHz-13GHz source
 * New hardware file for AWG - Keysight M3202A 1GS/s 4-channel PXIe AWG
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -25,7 +25,9 @@ This can be used to specify the axis labels for the measurement (excluding units
 * Bug fixes and support for SMD12 laser controller
 * For SMIQs added config options to additionally limit frequency and power. Added constraint for SMQ06B model.
 * Added live OMDR functionality to only calculate the average signal over a limited amount of scanned lines
-* New hardware file for Microwave source - Anritsu MG3691C has been added.
+* New hardware file for Microwave source - Anritsu MG3691C with SCPI commands has been added.
+* **Potential Config Change:** Hardware file for mw_source_anritsu70GHz.py with class MicrowaveAnritsu70GHz was changed to file mw_source_anritsu_MG369x.py with class MicrowaveAnritsuMG369x to make it universal. Also hardware constraints are set per model.
+* Lock-In functionality was added to the ODMR counter and implemented for the NI-Card. All other hardware and interfuse with ODMRCounterInterface were updated.
 * New hardware file for Microwave source - WindFreak Technologies SynthHDPro 54MHz-13GHz source
 * New hardware file for AWG - Keysight M3202A 1GS/s 4-channel PXIe AWG
 * Add separate conda environments for windows 7 32bit, windows 7 64bit, and windows 10 64bit. 
--- a/hardware/national_instruments_x_series.py
+++ b/hardware/national_instruments_x_series.py
@@ -1717,7 +1717,11 @@ class NationalInstrumentsXSeries(Base, SlowCounterInterface, ConfocalScannerInte
             return True, np.array([-1.])
 
         # check if length setup is correct, if not, adjust.
-        odmr_length_to_set = length * self.oversampling * 2 if self._odmr_pulser_daq_task else length
+        if self._odmr_pulser_daq_task:
+            odmr_length_to_set = length * self.oversampling * 2
+        else:
+            odmr_length_to_set = length
+
         if self.set_odmr_length(odmr_length_to_set) < 0:
             self.log.error('An error arose while setting the odmr lenth to {}.'.format(odmr_length_to_set))
             return True, np.array([-1.])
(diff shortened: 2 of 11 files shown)


hardware/national_instruments_x_series.py
```diff
@@ -33,6 +33,11 @@
 from interface.confocal_scanner_interface import ConfocalScannerInterface
 
 
+# Define the python callback function
```
drogenlied: This does nothing, please remove.

hardware/odmr_counter_dummy.py
```diff
@@ -48,6 +49,10 @@ class ODMRCounterDummy(Base, ODMRCounterInterface):
     _clock_frequency = ConfigOption('clock_frequency', 100, missing='warn')
     _number_of_channels = ConfigOption('number_of_channels', 2, missing='warn')
 
+    _pulse_out_channel = 'dummy'
```
drogenlied: No class variables!

interface/odmr_counter_interface.py
```diff
@@ -29,6 +29,10 @@ class ODMRCounterInterface(metaclass=InterfaceMetaclass):
     _modtype = 'ODMRCounterInterface'
     _modclass = 'interface'
 
+    _pulse_out_channel = 'dummy'
```
drogenlied: No class variables!

interface/odmr_counter_interface.py
```diff
@@ -102,4 +106,32 @@ def get_odmr_channels(self):
 
         @return list(str): channels recorded during ODMR measurement
         """
-        pass
\ No newline at end of file
+        pass
+
+    @property
```
drogenlied: Abstract methods only.
```
@property 
@abstractmethod 
```

should work according to documentation https://docs.python.org/3/library/abc.html#abc.abstractmethod

logic/interfuse/odmr_counter_microwave_interfuse.py
```diff
@@ -44,6 +44,11 @@ class ODMRCounterMicrowaveInterfuse(GenericLogic, ODMRCounterInterface,
     slowcounter = Connector(interface='SlowCounterInterface')
     microwave = Connector(interface='MicrowaveInterface')
 
+    _pulse_out_channel = 'dummy'
```
drogenlied: No class variables, either ConfigOption or StatusVar or think about not storing this in the hardware or interfuse module.

hardware/national_instruments_x_series.py
```diff
@@ -1562,6 +1572,19 @@ def set_up_odmr(self, counter_channel=None, photon_source=None,
...
             daq.DAQmxStopTask(self._scanner_clock_daq_task)
 
+            if self._lock_in_active:
+                ptask = daq.TaskHandle()
+                daq.DAQmxCreateTask('ODMRPulser', daq.byref(ptask))
+                daq.DAQmxCreateDOChan(ptask, self._pulse_out_channel, "ODMRPulserChannel", daq.DAQmx_Val_ChanForAllLines)
+
+                # Convert the python function to a C function callback
+                # The name is defined in DAQmxTypes
+                DoneCallback = daq.DAQmxDoneEventCallbackPtr(_DoneCallback_py)
```
drogenlied: This does nothing, please remove.

hardware/national_instruments_x_series.py
```diff
@@ -1655,22 +1716,48 @@ def count_odmr(self, length=100):
...
             if len(self._scanner_ai_channels) > 0:
                 daq.DAQmxStartTask(self._scanner_analog_daq_task)
         except:
             self.log.exception('Cannot start ODMR counter.')
-            return np.array([-1.])
+            return True, np.array([-1.])
+
+        try:
+            if self._odmr_pulser_daq_task:
+                pulse_pattern = np.zeros(self._oversampling*2, dtype=np.uint32)
```
drogenlied: Maybe a comment about what this pattern is doing?

hardware/national_instruments_x_series.py
```diff
@@ -1562,6 +1572,19 @@ def set_up_odmr(self, counter_channel=None, photon_source=None,
             # otherwise, it will be low until task starts, and MW will receive wrong pulses.
             daq.DAQmxStopTask(self._scanner_clock_daq_task)
 
+            if self._lock_in_active:
+                ptask = daq.TaskHandle()
+                daq.DAQmxCreateTask('ODMRPulser', daq.byref(ptask))
+                daq.DAQmxCreateDOChan(ptask, self._pulse_out_channel, "ODMRPulserChannel", daq.DAQmx_Val_ChanForAllLines)
```
drogenlied: This can be written in a way that does not use/overwrite all 32 channels of Port0.
Only the 2 lines that are needed should be reserved and written.

documentation/changelog.md
```diff
@@ -25,7 +25,9 @@ This can be used to specify the axis labels for the measurement (excluding units
 * Bug fixes and support for SMD12 laser controller
 * For SMIQs added config options to additionally limit frequency and power. Added constraint for SMQ06B model.
 * Added live OMDR functionality to only calculate the average signal over a limited amount of scanned lines
-* New hardware file for Microwave source - Anritsu MG3691C has been added.
+* New hardware file for Microwave source - Anritsu MG3691C with SCPI commands has been added.
+* **Potential Config Change:** Hardware file for mw_source_anritsu70GHz.py with class MicrowaveAnritsu70GHz was changed to file mw_source_anritsu_MG369x.py with class MicrowaveAnritsuMG369x to make it universal. Also hardware constraints are set per model.
```
drogenlied: Not just potential.

hardware/national_instruments_x_series.py
kay-jahnke: done

interface/odmr_counter_interface.py
kay-jahnke: that works.

hardware/national_instruments_x_series.py
kay-jahnke: still needs to be tested

logic/interfuse/odmr_counter_microwave_interfuse.py
kay-jahnke: deleted, as the abstractmethods only have a pass.

hardware/national_instruments_x_series.py
kay-jahnke: done.

drogenlied:
One more thing: the set_ext_trigger() function in the microwave source needs the timing of the microwave switching frequency in case of oversampling, not the sample clock frequency.

drogenlied:
One more: odd number of oversampling steps does not work.

Neverhorst:
As I see it someone needs to test the normal (non-lock-in) functionality to check if something has changed, is that right?
The lock-in functionality can only be tested by you guys anyway...

kay-jahnke:
I think we caught all the bugs now. Testing on the LP_Optics setups showed good results.

@Neverhorst maybe you can test the normal ODMR again on a confocal, just to be sure.

PR #460, opened 2018-12-11
Added Properties to PredefinedGeneratorBase

Neverhorst:
## Description
Added the two `SequenceGeneratorLogic` methods to analyze PulseBlockEnsembles and PulseSequences as properties to `PredefinedGeneratorBase`:

- `analyze_block_ensemble`
- `analyze_sequence`

Also added the methods to save PulseBlocks/PulseBlockEnsembles/PulseSequences:

- `save_block`
- `save_ensemble`
- `save_sequence`

It should be noted that if you want to make use of the analyze-methods inside a predefined generate method you would need to explicitly save the corresponding blocks/ensembles beforehand.

## Motivation and Context
If you want to do some calculations inside the predefined methods that require exact knowledge of the sampled waveform/sequence (e.g. matching an exact waveform length) you need to be able to access these analysis methods. You could also do that by using the reference to `SequenceGeneratorLogic` but this should be avoided since it is sensitive to changes, thus the new properties.

## How Has This Been Tested?
On dummy Win8.1 x64

## Types of changes
- [] Bug fix (non-breaking change which fixes an issue)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/documentation/pulsed/how_to_add_predefined_methods.md
+++ b/documentation/pulsed/how_to_add_predefined_methods.md
@@ -43,7 +43,7 @@ dictionary containing a variety of information about the to-be-sampled waveform/
 Refer to the documentation in `SequenceGeneratorLogic` to learn more about this dictionary.
 If you want to analyze the currently created `PulseBlockEnsemble`/`PulseSequence` within the 
 generate method, you need to explicitly save the corresponding `PulseBlocks` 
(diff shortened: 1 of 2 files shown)


kay-jahnke:
Do you want to add the way of how to use these methods to any documentation or changelog?

Neverhorst:
Good point @kay-jahnke. 
Added a few lines in the corresponding how-to.

PR #469, opened 2019-01-21
Joystick development

alrik-durand:
## Description
I've added interface and logic to use a controller in Qudi. I've created the hardware module for the xbox controller (based on xinput) and a connector to act on the confocal logic based on controller events.

## Motivation and Context
This change is not required, but it's kind of cool, and for people who spend a lot of time moving the confocal around it's very usefull.

## How Has This Been Tested?
This has been tested on Windows 10 with a controller like this one :  [link](https://www.microsoft.com/en-us/p/xbox-wireless-controller/8vcw8gln9vrf), with wire and wireless (with standart bluetooth usb dongle).
Looks good !

## Types of changes

## Checklist:


Neverhorst:
That is a very cool idea, thank you.

However this needs a bit more testing. I have tried with the dummy config for a couple of minutes and this is what I found so far:

1. The huge bunch of signals (one for each button/axis) must be eliminated. Usually control inputs are handled with a bunch of flags per frame encoded in some way indicating which buttons are currently pressed. You can simply emit a single signal and use it to hand over some container encoding the controller state.

2. When using the analog sticks on the controller and releasing them while keeping the interlock button pressed, the confocal position keeps slowly drifting (even if you set a large axis threshold in the config).

3. In `xinput.py` line 96 the scaling values need to be handled in a more general way. If I understand them correctly these values are controller-specific and as such can not be hard-coded.

4. It took me quite some time to figure out the controls. Before this PR can be merged you need to write a detailed "how-to" about the usage of the controller and add documentation about the implementation concept (since a non-default input device is something entirely new in qudi).

5. The code needs more comments/documentation

6. The interlock button concept is a bit cumbersome. I understand that you want to be able to lock the cursor movement but to my mind it would be more user-friendly to toggle the controls. (switch movement on/off by a button push).

7. The whole concept of polling the controller state with a QTimer needs to be carefully investigated. Usually you want to poll the controller state with a very high frequency (to register short button push actions) but act on the inputs in a more controlled (filtered) and slow way.
Regarding this point it may be interesting to explore the possibilities of the Qt `QGamepad` class. Though I'm not sure we can use this class in the current state of qudi.

8. It would be quite nice to have the ability to trigger position optimization from the controller.

All in all it already works quite well but I'm a bit concerned about the way the controller state is polled and how the events are handled/handed over to confocal.

alrik-durand:
Hello again !

It took me a while but I've redesign most of the logic. I think this is way cleaner now !
I've also added some doc on how to use it, and the optimizer ;)

I had to kept the Qtime because time.sleep() is synchronous and so kept crashing Qudi.
The PR is now ready for another review !

Neverhorst:
Thank you @alrik-durand. 
I will review that as soon as possible but it will probably take some time.

Neverhorst:
So, I have been thinking about gamepad/joystick integration and discussed also with @drogenlied.

What bothers me in the proposed approach is that the input device is a qudi hardware module.
User interface devices should only ever be used to control GUI elements, thus be only handled by GUI modules. Since GUI modules must not connect directly to hardware modules the current approach causes logic modules to handle user input, which is very messy and contradicts the strict separation between the functional program and user interface.
Another downside is the cluttering of many logic modules with gamepad-handling code that is rarely used and decreases readability (remember qudi is for non-programmers).

Let's take a step back...
Input devices should be controlling GUI elements only. This is provided by Qt via key press events and mouse tracking/click events. For C++ Qt offers the `QtGamepad` addon. However this addon is unfortunately not available for PyQt5 and PySide2 (Qt for Python). Hopefully it will be in the future but this is not in sight, yet.
So, how can we integrate gamepad input events into the Qt event handling mechanism?
One way I came across is `QtTest`. It can be used to inject any desired GUI event. Using this package one could map XInput events to key press and mouse events.
I tested this in: https://github.com/Ulm-IQO/qudi/blob/gamepad_sandbox/core/gamepad.py
(It's just a tech demo and is far from useable)
Turns out this works not so bad and imposes no additional complexity to qudi modules.
So key mapping should not be an issue (if hidden properly in GUIBase class and core) and could for example be configurable via config.
What remains are analog axis inputs.
I currently see 3 possible roads to go down for analog inputs:

1. Providing a QtSignal to the GUIBase class that carries analog axis events. This approach is problematic since rapid signal firing must not stall the main event loop. One would also need some kind of signal proxy that can dump/ignore signals in an orderly fashion if fired too fast to process (similar to the Qt event system... maybe even try to tap into that). GUI modules would also need to provide extra specialized slots for analog axis inputs.

2. Investigate possible uses of `Qt3DInput` which is available for Python. This could potentially be used (not properly investigated by me) to handle analog axis inputs. The buttons could then be mapped like in the previous case.

3. Map gamepad buttons to keyboard keys (like in 1.) and 3 analog axes to mouse movements (x, y and wheel). This would have the advantage of GUI elements not needing to distinguish between joystick/gamepad and mouse. This would definetely be the most universal/compatible approach.
Only downside is that not all capabilities of a modern gamepad can be used. (XBox controller has 6 analog axes)
In case of the confocal GUI one would then be able to activate direct mouse control for a certain scan widget by some trigger (hotkey or doubleclick or ...) and the crosshair would follow the mouse just by moving it around. If one stick of the gamepad is mapped to the mouse, this works just the same.

So as you can see there are more or less elegant ways to deal with gamepad inputs without shaking qudi's foundations. However this is no easy feature to implement properly. While it would be cool to use joysticks or gamepads to control GUIs in qudi, it remains a non-essential, edge-case feature.
That being said I can not spare the time right now to actively help implementing that. I'm certainly happy if you would follow this path further down and will of course review new implementations as well as share my insights on the topic.
The way it is currently implemented will unfortunately not be merged into master due to the reasons mentioned in the beginning.

alrik-durand:
Hi,
First thank you for all the effort you put in reviewing this PR.

I'll take the time to look into the different options you propose, I'm sure we can find something we can all agree on. 

The reason I did this work was because I wanted to be able to control a stepper device easily, something Qudi is not currently doing. The connection to scanner logic was more a demo than a really useful feature, as the GUI is already great for that. (I can guarantee that for the steppers, it's very nice !)
I admit I did not look enough into QtGamepad, as I read it wasn't available for PyQt. I'll look it up.

alrik-durand:
Hi,
This PR has been open for quite some time and with the experience I now realize that I made a mistake in developing/presenting this feature as a general, multi-function device.
The benefit of a wireless control for confocal module is negligible and we never used it. The real usage we have for it is to control one and only one module, in our case the sample position via attocube open-loop steppers (via motor interface).
On a day to day basis, it's very useful because the controller does not depend on the current window and is similar to the 1-axis joystick on the stepper controller.
Therefore, I'm not interested in developing additional gamepad control to every GUI, but stay close to the current architecture.

Now that this is clear, should I :
1. Close this PR and use these module only internally in our lab ?
2. Improve this the current PR so that it is easily accessible for everyone ?
3. Something else ?

Neverhorst:
Well, as I mentioned in earlier comments implementing gamepad support in a non-hacky way is quite a big task.
So I completely understand your standing on this.

I think the joystick hardware and logic module are a good starting point for people who want to revisit this topic in the future. As such it would be a waste to just throw it away.
On the other hand I do not want this in master because people would start interfacing their modules with it and we do not want that unless it is implemented properly.

So my best bet is on cerating a branch in this repo and pull the joystick modules in there (just the neccessary ones, not the ones interfacing other logic modules). This would enable people to start from there once someone is willing to implement this.

Neverhorst:
See branch [gamepad_sandbox2](https://github.com/Ulm-IQO/qudi/tree/gamepad_sandbox2)

PR #470, opened 2019-01-21
Sweeps in pulsed measurement

alrik-durand:
## Description
I've modified the fast_counter_interface to get the number of sweeps, and updated all the hardware with a dummy function. I've also updated the logic to take this into account, espacially in the stop and restart features. I've finally taken paused time into account in elapsed time.

## Motivation and Context
The sweeps feature was shown in the GUI but not implemented yet.
The elapsed time should not increase when the acquisition is paused.

## How Has This Been Tested?
I've tested this on the dummies and on a real setup with a fastcomtec 7887.

## Types of changes

## Checklist:

### DIFF
--- a/logic/pulsed/pulsed_measurement_logic.py
+++ b/logic/pulsed/pulsed_measurement_logic.py
@@ -124,7 +124,7 @@ class PulsedMeasurementLogic(GenericLogic):
         self.__analysis_timer = None
         self.__start_time = 0
         self.__elapsed_time = 0
-        self.__elapsed_sweeps = 0  # FIXME: unused
+        self.__elapsed_sweeps = 0
 
         # threading
         self._threadlock = Mutex()
@@ -821,9 +821,8 @@ class PulsedMeasurementLogic(GenericLogic):
 
                 # stash raw data if requested
                 if stash_raw_data_tag:
-                    #self.log.info('Raw data saved with tag "{0}" to continue measurement at a '
-                    #              'later point.'.format(stash_raw_data_tag))
-                    self._saved_data[stash_raw_data_tag] = {'raw': self.raw_data.copy()}
+                    self._saved_data[stash_raw_data_tag] = {'raw': self.raw_data.copy(),
+                                                            'sweeps': self.__elapsed_sweeps}
                 self._current_saved_data_tag = None
 
                 # Set measurement paused flag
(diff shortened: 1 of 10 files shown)


alrik-durand:
Thanks for all your feedbacks, it helps a lot !
I will try to implement all that

alrik-durand:
1. I can try to implement it, but I'm not sure what are the possible parameters and how the logic should treat them

2. I'll remember that, I edited the files even if this will not be kept

3. I asked myself the question when writing it, but I felt like changing the time of start was cheating a bit, if you think it's cleaner I'll edit it no question ask !

4. I was wondering what should be prefered : in on_activate, in __init__ or in class ? 

5. Done

6. Will do ! (I mean, not do!)

7. I understand now, I can't fix this easily because of the way my code handle sweeps right now

8. Ok !

9. Will do when everything is worked out.

Neverhorst:
1. Well, the only two keys in the proposed `info_dict` will be `'elapsed_sweeps'` and `'elapsed_time'`. The logic can check these keys and if they are missing or negative or None then it should calculate the elapsed_time as usual on its own.

3. Do it! ;)

4. These three locations differ vastly in what they do. As a rule of thumb you should never add something to the class itself (except `StatusVar`, `ConfigOption`, `Connector` and `QtCore.Signal`).
All the variables you need should be declared in the `__init__` but this will be just executed the first time you load this module. In qudi you are able to reload (deactivate and activate) a certain module during runtime. So the deactivation should contain everything you need to perform in order to properly clean up after you (close device connections, terminate measurements etc.). The activation will be called (in contrast to the `__init__`) EVERYTIME you activate/reload the module. So in order to have a clean start after reloading the module you need to re-initialize all necessary parameters inside `on_activate`. If you do not comply to this scheme you could end up with corrupt data or settings from previous module sessions.

7. Care to elaborate?

alrik-durand:
This works on dummy, I haven't changed any hardware module.

Right now the logic support the old interface so that it won't be a breaking change.

logic/pulsed/pulsed_measurement_logic.py
```diff
@@ -1142,29 +1157,47 @@ def _get_raw_data(self):
...
+                                                 info_dict with keys 'elapsed_sweeps' and 'elapsed_time'
         """
         # get raw data from fast counter
         fc_data = netobtain(self.fastcounter().get_data_trace())
+        if type(fc_data) == tuple and len(fc_data) == 2:  # if the hardware implement the new version of the interface
+            fc_data, info_dict = fc_data
+        else:
+            info_dict = {'elapsed_sweeps': None, 'elapsed_time': None}
+
+        if type(info_dict) == dict and 'elapsed_sweeps' in info_dict.keys() and info_dict['elapsed_sweeps'] is not None:
```
Neverhorst: Please avoid `type(some_var) == dict` or similar for these kind of checks.
It's better to use `isinstance(some_var, dict)` since this will also return `True` in case of derived types (e.g. `OrderedDict`).
You can also do something like `isinstance(some_var, (list, tuple))` to check for both.

Instead of 
`'elapsed_sweeps' in info_dict.keys() and info_dict['elapsed_sweeps'] is not None` 
you could do 
`info_dict.get('elapsed_sweeps') is not None`. This is more efficient, shorter and easier to read.

logic/pulsed/pulsed_measurement_logic.py
```diff
@@ -1142,29 +1157,47 @@ def _get_raw_data(self):
         """
         Get the raw count data from the fast counting hardware and perform sanity checks.
         Also add recalled raw data to the newly received data.
-        :return numpy.ndarray: The count data (1D for ungated, 2D for gated counter)
+        :return tuple(numpy.ndarray, info_dict): The count data (1D for ungated, 2D for gated counter) and
```
Neverhorst: please write `@return numpy.ndarray:`

alrik-durand:
Thanks for your remarks, I've updated accordingly !

PR #476, opened 2019-01-29
Fix sequence flags

kay-jahnke:
Only one flag could be selected for sequence steps in the sequence generator. This has now been updated to a checkbox for each flag, similar to the digital channels of the PulseBlocks.

## Motivation and Context

A sequence step might have two or more flags activated at the same time.

## How Has This Been Tested?

Win 7 64x on dummy. Running from GUI as well as from script.

grafik

## Types of changes

## Checklist:


kay-jahnke:
I will try finding an 70k AWG next week.

I did find one unfortunate thing: When loading an old sequence (stored with the old master) the SequenceStep throughs a KeyError, as the flags are no lists. This causes the table to have the right number of lines but with no content. Clearing the table does also not work anymore.

Not sure what goes wrong here.

PR #479, opened 2019-02-06
Fix analyse sequence

kay-jahnke:
This PR actually get the analyse sequence function working. Therefore it first analyses all the ensembles in the sequence and then compiles the complete information for the sequence.

## Description

Get the correct sequence information.

## Motivation and Context

Make sequences usable for example by using gated counting that relies on the information generated at sequence generation.

## How Has This Been Tested?

Tested with dummy.

## Types of changes

## Checklist:


Neverhorst:
I've had some problems with the `ungated_gated_conv_deriv` method. I adjusted it now to the recent changes in `analyze_sequence` and cleaned up the code a bit. Can someone with an ungated counter check if everything is working as intended?

PR #485, opened 2019-02-19
Sequence channel flank detection

Neverhorst:
## Description
Improved `SequenceGeneratorLogic.analyze_block_ensemble` and `SequenceGeneratorLogic.analyze_sequence` to be more efficient and most importantly handled special cases of laser_on flag / digital channel rising/falling flank detection in sequences.

## Motivation and Context
The recently introduced expansion of the `analyze_sequence` return dict was incomplete. Special cases of the digital flank detection when transitioning waveforms in sequences were ignored and are now properly handled. 

## How Has This Been Tested?
On dummy Win8.1 x64
Tested using a notebook that was creating random pulse sequences and compared the "brute force" approach (walking each `PulseBlockElement` as it would be played) result to the now introduced `analyze_sequence` return values.

## Types of changes

## Checklist:

### DIFF
--- a/logic/pulsed/pulse_extraction_methods/basic_extraction_methods.py
+++ b/logic/pulsed/pulse_extraction_methods/basic_extraction_methods.py
@@ -354,6 +354,20 @@ class BasicPulseExtractor(PulseExtractorBase):
         # get laser rising and falling bins
         laser_rising_bins = self.sampling_information['laser_rising_bins']
         laser_falling_bins = self.sampling_information['laser_falling_bins']
+
+        # Sort out trailing or leading incomplete laser pulse
+        while len(laser_rising_bins) != len(laser_falling_bins):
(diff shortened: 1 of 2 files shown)


kay-jahnke:
fixed a small bug for just laser on. Please check again.

Also now (as before) the laser is not detected if the ensemble ONLY contains laser blocks. Is this intended?

Neverhorst:
Thank you.
Yes, this is intended behaviour. If the entire playback of the sequence or waveform never switches the laser_channel state, then all of the pulsed tools will not work properly.
There is however an improvement: Contrary to the current master, the laser channel transition will now be properly registered when you have two waveforms in a sequence that have a static laser channel state (one constant low, the other constant high). In that case both waveforms (or `PulseBlockEnsembles`) will return zero laser pulses but the `PulseSequence` containing them will return >0 laser pulse

PR #487, opened 2019-02-20
update

SimonSchmitt88:
Added a few methods and implemented the return of elaspsed sweep 

## Motivation and Context

Update

## Types of changes

## Checklist:

### DIFF
--- a/hardware/fastcomtec/fastcomtecmcs6.py
+++ b/hardware/fastcomtec/fastcomtecmcs6.py
@@ -162,8 +162,6 @@ class BOARDSETTING(ctypes.Structure):
 class FastComtec(Base, FastCounterInterface):
     """ Hardware Class for the FastComtec Card.
 
-    stable: Jochen Scheuer, Simon Schmitt
-
     Example config for copy-paste:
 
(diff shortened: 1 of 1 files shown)


kay-jahnke:
In the function get_data, why do you first get the runtime but then return a value of None?

SimonSchmitt88:
Thanks for reviewing @kay-jahnke. This time is the running time of the fastcomtec and not the running time of the experiment. Depending on the loading duration of the pulser there can be a minutes difference. After noticing that, I changed the return value to None.

Neverhorst:
@SimonSchmitt88: Would it help to change pulsed measurements to always start the pulser before starting the counter?

Neverhorst:
What's the status here, @SimonSchmitt88, @kay-jahnke, @NicolasStaudenmaier, @MathiasMetsch?
Is this branch even in use on some setup or has it become obsolete and another solution has emerged?

SimonSchmitt88:
I don't think that this branch is in use at the moment. I close this PR and will have a look if it is worth to create a new PR soon.

PR #494, opened 2019-03-18
Poimanager Overhaul

Neverhorst:
## Description
Cleanup and debug of the POI manager.

Overall performance improvement.
Better programming interface when working with `PoiManagerLogic`.
ROI scan image will now move according to sample shift.
Size of the POI marker circles will now correspond to optimizer scan window size.
Removed buggy/unreliable ROI reorientation functionality.
Automatic periodic refocus is more robust and will not raise hell if optimizer/scanner logic is busy when next refocus is due.

## Motivation and Context
The POI manager as it was had a lot of "sharp edges" and bugs. Re-implemented its functionality in a clean and more qudi-conform way. Also improved the GUI performance when working with a large number of POIs.

## How Has This Been Tested?
On Win8.1 x64 dummy config

## Types of changes

## Checklist:


kay-jahnke:
One more when closing down qudi (or deactivating the logic):
grafik
Remark: this does not happen if I generate a new ROI, but I can choose a name (including ,.;:""''<>), that crashes the POI manager.

Neverhorst:
@kay-jahnke: I think you might have a different problem here. The POI markers don't look like they are supposed to and I can not reproduce your errors. Are you sure you did not confuse/mix up code bases? Maybe also delete app-status before testing.

liuyanulm:
Tested on my office computer, with dummy data.
No error occurred.
Nice marker size design. Thanks!

kay-jahnke:
You are correct. I also cannot reproduce the errors described above, after I deleted my app_status.

I did however find a new error, when I try to reload the PoiManager logic from the manager gui:
grafik
This is with a clean app_status on Win7 64x, dummy.

drogenlied:
I can confirm the latest error on Linux.

Neverhorst:
Thank you all for testing.

I can confirm the latest error (during reload of the logic module) and I have fixed it now. 

Also the previous error that was reported during deactivation of the module has been confirmed and only happens if you have loaded a ROI from disk before deactivating the module.
Apparently the check `isinstance(mystring, str)` equals `True` for `type(mystring) == numpy.str_` but  `ruamel.yaml` is unable to represent the numpy string. All setters for names have been adjusted to explicitly typecast to `str`.

Also added legacy support for old ROI files. The name of the ROI may contain an additional POI tag however but this is ok I think.

drogenlied:
Additional small things:
  * POIs have handles for resizing, should they have these?
  * Passing a numpy array into StatusVar creates a file, it would be nice if this happens only for big arrays like images and not for a 3 element position array.

Neverhorst:
Regarding your remarks @drogenlied:

- This is the same weird behavior as before with what @kay-jahnke mentioned in the first comment. I can not reproduce this issue somehow. The custom class `PoiMarker` in `gui/poimangui.py` is redefining the method `_addHandles` and thus avoiding the creation of any handle. In my case I have just a circle with no handle and a label.

- Good point. Fixed that for individual POI positions. The ROI history and the scan image are still saved in a `.npz` file. But this is ok I think (max. 2 numpy arrays for `PoiManagerLogic`) . This issue makes me wondering however why there is not a single `.npz` file per module. The numpy format allows several arrays to be stored in a single file. This would avoid file-spamming and is more efficient/clean.

Neverhorst:
Guys, I just added a feature that I need desperately.
There is a new tool in the toolbar now - the "POI selector".
While this tool is enabled, mouse clicks inside the ROI scan image will create new POIs at this position as long as you are not clicking on an already present POI marker.
This new feature should not mess with the stuff you already tested since it is a pure GUI change.

kay-jahnke:
The new feature is quite nice. The only problem is, if you don't hit the POI properly the sample will be shifted by your click error: the way around is, to go to POI, refocus crosshair and then move POI anchor. That should do in for now.

With deleting the app-status I still get weird erros:
For me the POI names get wrapped by b'...' and that multiple times if I reload the ROI more than once.
Also the hande for the resize I can confirm (you find it if you zoom in quite a lot). Also the naming handle moves with the sizes of the marker, but the name does not. That looks weird.

grafik

Neverhorst:
Ok the behaviour is extremely weird and is completely different from what I can see. Let's discuss that in person.

I understand your concerns with the sample shift but I think you are using the POI manager differently. The only difference with clicking is that you do not have to move the crosshair, switch back to POI manager and add the position as POI and then repeat.
If you add POIs by optimizing the position for every spot before adding it as POI you accumulate errors the longer it takes between the first and the last POI added to the ROI. If you want your POI positions to perfectly correspond to a confocal scan snapshot you must quickly add the POIs by clicking the spots in the image (zoom in if you like for increased accuracy). After that you optimize a single reference POI in that ensemble of POIs and you have matching distances (good enough to jump on them with a single refocus). I can explain in person... that is tedious to describe.

drogenlied:
Some more thoughts about the nametag and POI Selector tools:

The nametag text field should probably be named 'POI prefix'.
Clearing the text input leads to new POIs being created with the old value that is not there any more, as the validator probably fails but the text field does not reflect that.

The `POI Selector` tool should probably be named `Add POIs` and show a crosshair cursor when hovering over the confocal image view.
It could also be named `New POI` if we didn't have that name already.

The `New POI` button should probably be named `POI from Confocal` to be more clear about what happens in contrast with the other button.

drogenlied:
I think I found the source of the handles problem, pyqtgraph-qo 0.10.0.5 does not show handles while stock pyqtgraph 0.10.0 does and older versions of pyqtgraph-qo may as well.

Neverhorst:
Thanks for pointing out the validator bug. Will fix that.

Regarding the GUI design issues I had the exact same thoughts but was too lazy to create a new icon so far. But will do...

Ok but if pyqtgraph-qo 0.10.0.4 shows the problem why does a clean environment install fix it if this is the version installed by default?

drogenlied:
Just tested some versions, 0.10.0.2 seems to be the last version to show handles 0.10.0.3 and 0.10.0.4 are fine, at least on my laptop.
This could also be a pyqtgraph and Qt version problem. I have 5.11, the conda environment should have 5.9 while old installs could be 5.6 or 5.7.

Neverhorst:
How about that, @drogenlied?

Do you really think it's necessary to change the cursor to something else when hovering over the ROI scan image? I would need to do a lot of reading on how to do that properly and as far as I know we have no similar behavior in qudi so far. But I agree that it would be nice though...

Neverhorst:
Well that was easier than I expected. Cursor is changing now.

In order to implement the cursor and the range checks in a more elegant and "Qt-like" fashion one would need to create a custom `PlotWidget` sub class and redefine mouse events.

drogenlied:
Good stuff.

liuyanulm:
errors occurred during activation of Poimanager Gui.

1. ValueError: too many values to unpack (expected 2)

2. AttibuteError: 'PoiMangerLogic' object has no attribute 'roi_scan_image'

Tested on Windows 10 with default.cfg.

Neverhorst:
> errors occurred during activation of Poimanager Gui.
> 
> 1. ValueError: too many values to unpack (expected 2)
> 2. AttibuteError: 'PoiMangerLogic' object has no attribute 'roi_scan_image'
> 
> Tested on Windows 10 with default.cfg.

These errors should have been fixed already. Please check out the latest version of this branch for testing.

liuyanulm:
> > errors occurred during activation of Poimanager Gui.
> > 
> > 1. ValueError: too many values to unpack (expected 2)
> > 2. AttibuteError: 'PoiMangerLogic' object has no attribute 'roi_scan_image'
> > 
> > Tested on Windows 10 with default.cfg.
> 
> These errors should have been fixed already. Please check out the latest version of this branch for testing.

You are right. I have updated it, and now it works out fine. A thumb-up

PR #497, opened 2019-03-28
Scanimage plotwidget improvement

Neverhorst:
## Description
Introduced some new sub-classes of _pyqtgraph_ widgets/items in order to handle coordinate transformations (between items, scene and view) for mouse click and drag events within the new sub-classes.
New classes can be found in `qtwidgets.scan_plotwidget.py`.
Debugged the zoom functionality in confocal while at it.
Affected modules are `ConfocalGui`, `PoiManagerGui` and `MagnetGui`.
Added blink correction for confocal and poimanager scan images. So far this is only for display in the GUI; raw data is not altered. Toggle this filter from the "view" menu.
Added the blink correction to `core.utils.filters`.

## Motivation and Context
This reduces confusing "deep-pyqtgraph" code within the qudi GUI modules. Also reduces code duplication and is a better basis to add functionality to these widgets later on.
The direct alteration of the _Qt_ class event callbacks to add functionality should always be the way to go since it's how _Qt_ is intended to be used... although _pyqtgraph_ violates this concept quite often.

## How Has This Been Tested?
Pure GUI change, so just on dummy Win8.1 x64

## Types of changes

## Checklist:


Neverhorst:
Ah, just saw that I forgot docstrings in the classes and some methods. Please stand by...

Neverhorst:
Ok docstrings should be up-to-date now. Also added first version of blink correction (display only, no messing with data).

drogenlied:
```
  File "/home/jan/build/qudi/gui/confocal/confocalgui.py", line 1412, in refresh_refocus_image
    self._optimizer_logic._initial_pos_y + 0.5 * self._optimizer_logic.refocus_XY_size))
TypeError: set_image_extent() takes 2 positional arguments but 3 were given
```
This happens when I optimize the position in confocal.

Neverhorst:
Sorry, I already fixed that bug but I forgot to commit.

drogenlied:
I noticed one more thing: at 100x100 µm scan area and standard refocus area, it is nearly impossible to grab and move the cursor.

Neverhorst:
Yes, I noticed that. There is an open question how this should be handled:

Before you had the ROI drawn (the small rectangle you see now) and in addition you had a slightly bigger translation handle (scaling with viewport) resulting in two overlapping rectangles, which is not so nice in my opinion. Also colors (pens) for the handles can not be properly controlled.
On the other hand you can now also drag the horizontal and vertical lines.

You can also overcome this issue (same as before) by unchecking "adjust cursor size to optimizer range" in the settings dialogue. Then the rectangle will scale with zooming and the scan range to be always 5% of the shown image size. But of course then you don't have the optimizer window feedback.

If this is desired I can easily reintroduce the redundant translation handle but then you always have two boxes. (regardless of the checkbox state described above)

core/util/filters.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+
+import numpy as np
+from scipy.ndimage import minimum_filter1d, maximum_filter1d
+
+import logging
+logger = logging.getLogger(__name__)
+
+
+def scan_blink_correction(image, axis=1):
```
kay-jahnke: Please document what this does.

gui/confocal/confocalgui.py
```diff
@@ -1752,69 +1542,72 @@ def adjust_depth_window(self):
...
-            if view_y_max < cross_pos[0]:
-                horizontal_value = view_y_max - self.roi_depth.size()[1]
-
-        if view_z_min > cross_pos[1]:
-            vertical_value = view_z_min + self.roi_depth.size()[1]
-        if view_z_max < cross_pos[1]:
-            vertical_value = view_z_max - self.roi_depth.size()[1]
-
-        self.roi_depth.setPos([horizontal_value, vertical_value], update=True)
+    # def put_cursor_in_xy_scan(self):
```
kay-jahnke: please delete code not needed anymore.

qtwidgets/scan_plotwidget.py
```diff
@@ -0,0 +1,322 @@
...
+
+    sigMouseAreaSelected = QtCore.Signal(QtCore.QRectF)
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.zoom_by_selection = False
+        self.rectangle_selection = False
+        return
+
+    def toggle_selection(self, enable):
```
kay-jahnke: Why are you using toggle and not just `@property` and `@xyz.setter`?

Neverhorst: For two reasons mainly:

- Since this is a derived Qt5 widget I felt like it is better practice to stick with Qt-like implementations with function calls rather than pythonic. I just introduced some properties for read-only purposes (no setters). This is also what you would expect when using (py)Qt-native widgets.
- As Qt widget it is quite common to connect signals to this kind of methods. If you want to use the method as Qt slot, it's more convenient.

core/util/filters.py
Neverhorst: In general most methods and newly introduced functionality lacks documentation. I will add more anyways.

gui/confocal/confocalgui.py
Neverhorst: Done

core/util/filters.py
Neverhorst: Done

qtwidgets/scan_plotwidget.py
kay-jahnke: OK, I was genuinly just wondering.

At some point in the future (and when pyqtgraph gets there shit in order) we shoud look at best practices, but unifying the style is really something for the future.

PR #499, opened 2019-03-30
Process modifiers interfuses

alrik-durand:
## Description
I've added two interfuses for process_value and process_control interfaces.
Theses interfuses can use an interpolation function to convert the values in any way.

## Motivation and Context
Process values and control are useful for old kind of applications. But sometimes you're not interested in the value in itself but some function of it. Here is two examples that motivated me to create theses interfuses 
* Multiplying the value read from a power meter by a factor of 10 to monitor the power on the other side of a beam splitter
* Converting a simple setpoint power to a non linear voltage setpoint based on a calibration

## How Has This Been Tested?
I've checked with the demo hardware and the process_modifier_test.cfg demo config

## Types of changes

## Checklist:


config/example/calib/process_modifier_test.calib
```diff
@@ -0,0 +1,3 @@
+# X Y
```
Neverhorst: Please do not add a new folder _calib_ just for the example. Rather define a 1:1 interpolation as default value (no change in data) and do not enforce the load from file.

config/example/process_modifier_test.cfg
```diff
@@ -0,0 +1,57 @@
...
+    processdummy:
+        module.Class: 'process_dummy.ProcessDummy'
+
+logic:
+    process_value_modifier:
+        module.Class: 'interfuse.process_value_modifier.ProcessValueModifier'
+        connect:
+            hardware: 'processdummy'
+        calibration_file: 'config/example/calib/process_modifier_test.calib'
+        force_calibration_from_file: True
```
Neverhorst: see above.

config/example/process_modifier_test.cfg
```diff
@@ -0,0 +1,57 @@
...
+
+    process_control_modifier:
+        module.Class: 'interfuse.process_control_modifier.ProcessControlModifier'
+        connect:
+            hardware: 'processdummy'
+        calibration_file: 'config/example/calib/process_modifier_test.calib'
+        force_calibration_from_file: True
```
Neverhorst: see above.

logic/interfuse/process_control_modifier.py
```diff
@@ -0,0 +1,126 @@
+# -*- coding: utf-8 -*-
+
+"""
+This file contains the Qudi interfuse between a process control and a process control.
```
Neverhorst: Guess this is some copy/paste error?

logic/interfuse/process_control_modifier.py
```diff
@@ -0,0 +1,126 @@
...
+                                 'given. Please update calibration before using')
+        else:
+            self._update()
+
+    def on_deactivate(self):
+        """ Deactivate module.
+        """
+        pass
+
+    def _update(self, calibration=None):
```
Neverhorst: Please do not make the `update` method private. You will need to call this method from external modules/scripts to change the interpolation. Also please give it a more descriptive name ("update" is a bit too generic).

logic/interfuse/process_value_modifier.py
```diff
@@ -0,0 +1,103 @@
...
+                                 'given. Please update calibration before using')
+        else:
+            self._update()
+
+    def on_deactivate(self):
+        """ Deactivate module.
+        """
+        pass
+
+    def _update(self, calibration=None):
```
Neverhorst: see above

logic/interfuse/process_control_modifier.py
alrik-durand: Well no I don't think so. This interfuse connect to a process control and act like a process control.
Or am I missing something ?

Neverhorst: Ah well, yes you are right... it just looked weird.

PR #500, opened 2019-04-01
Sample ensemble only if needed

kay-jahnke:
So far all the ensembles needed in a sequence were first deleted and then sampled anew.
Now first a check is performed, if the ensemble already was sample before and it is only sampled if needed (so always when rotating wave is active).

## Description

Ensembles are only sampled if needed.
The analyze_sequence and analyze_ensemble functions are now available through pulsedmaster.
The aforementioned analyze functions can either be called with the appropriate objects directly or by the object's name.

## Motivation and Context

When an ensemble is created or changed, this ensemble should be newly sampled. But if the same ensemble is used by multiple sequences they should not be overwritten but rather all sequences should use the same ensembles.
This saves time on the one hand and several sequences can be pre-sampled and then just loaded to run experiments.

## How Has This Been Tested?

Tested on LP1 with AWG70k and Win 8.1.

## Types of changes

## Checklist:

### DIFF
--- a/logic/pulsed/sequence_generator_logic.py
+++ b/logic/pulsed/sequence_generator_logic.py
@@ -1152,7 +1152,7 @@ class SequenceGeneratorLogic(GenericLogic):
         PulseBlocks are actually present in saved blocks and the channel activation matches the
         current pulse settings.
 
-        @param ensemble: A PulseBlockEnsemble object (see logic.pulse_objects.py)
+        @param ensemble: A PulseBlockEnsemble object (see logic.pulse_objects.py) or the name of one
         @return: number_of_samples (int): The total number of samples in a Waveform provided the
                                               current sample_rate and PulseBlockEnsemble object.
                  total_elements (int): The total number of PulseBlockElements (incl. repetitions) in
@@ -1166,6 +1166,14 @@ class SequenceGeneratorLogic(GenericLogic):
                                              (in timebins; incl. repetitions) for each digital
                                              channel.
         """
+
+        if isinstance(ensemble, str):
+            ensemble = self.get_ensemble(ensemble)
+        elif not isinstance(ensemble, PulseBlockEnsemble):
+            self.log.error('Ensemble to analyze must either be of type PulseBlockEnsemble or the name of the ensemble.'
+                           'Returning empty dict')
+            return {}
+
         # Determine the right laser channel to choose. For gated counting it should be the gate
         # channel instead of the laser trigger.
         laser_channel = self.generation_parameters['gate_channel'] if self.generation_parameters[
@@ -1290,7 +1298,7 @@ class SequenceGeneratorLogic(GenericLogic):
         PulseBlocks are actually present in saved blocks and the channel activation matches the
         current pulse settings.
 
-        @param sequence: A PulseSequence object (see logic.pulse_objects.py)
+        @param sequence: A PulseSequence object (see logic.pulse_objects.py) or the name of one
         @return: number_of_samples (int): The total number of samples in a Waveform provided the
                                               current sample_rate and PulseBlockEnsemble object.
                  total_elements (int): The total number of PulseBlockElements (incl. repetitions) in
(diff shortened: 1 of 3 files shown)


logic/pulsed/sequence_generator_logic.py
```diff
@@ -1896,6 +1913,11 @@ def sample_pulse_sequence(self, sequence):
 
                 # Add created waveform names to the set
                 written_waveforms.update(waveform_list)
+            else:
```
Neverhorst: This `else:` clause must do much more than just taking the information from the ensemble instance as granted.
The case where the ensemble has been sampled with different pulse generator settings has to be handled. So each parameter in the `sampling_information` dict has to be sanity checked if it is still valid (active channels etc.).

logic/pulsed/sequence_generator_logic.py
```diff
@@ -1896,6 +1913,11 @@ def sample_pulse_sequence(self, sequence):
...
+                self.log.debug('Waveform already sampled: {0}'.format(name_tag))
+                ensemble_info = self.get_ensemble(name_tag).sampling_information.copy()
```
Neverhorst: The `ensemble_info` dict returned by `sample_pulse_block_ensemble` is not the same as `sampling_information` inside the ensemble instances. They share some information and it could even be that one is a subset of the other but they must be consistent either way. So do not simply copy but ensure that the dict contains exactly the same (and only the) information as would be returned by `sample_pulse_block_ensemble`.

SimonSchmitt88: I really like the idea, but is it possible to make it optional? It often happens to me that I run the same experiment several times with different parameters. So I want some of the ensembles which have been sampled previously to be overwritten. If I understand correctly I would then need to rename the new individual ensembles or I have to clear the old ones by hand or I have to generate the new ensembles in a seperate step...
Of course it would be optimal if I can decide before which ensembles to keep (and only sample if not present) and which to sample in any case, but I am not sure how to realise that

Neverhorst: @SimonSchmitt88:
Depending on HOW you introduce changes to an existing `PulseBlockEnsemble` this method would still sample it anew. 
Let's assume you have `myens` which is an instance of `PulseBlockEnsemble` and you want to set a new `PulseBlock` named `'myblk'` with 0 reps at index 1 by calling the following:
```python
myens[1] = ('myblk', 0)
```
If `myens` has been sampled before that, this setting of a new `PulseBlock` at index 1 will automatically trigger new sampling of the waveform. Actually it will every time you alter its items directly.

This is because you can handle these pulse objects like lists and every time the method `__setitem__` gets called (e.g. by the example above) it will delete the `sampling_information` dict inside the `PulseBlockEnsemble`. And this in turn will cause it to be assumed not sampled.
If you alter the `PulseBlock` instances directly, this will not be noticed.

I think the easiest solution for your scripts would be one of:

- Manually set `myens.sampling_information = dict()` if you alter `myens`
- Call the example above if you have altered `myblk`
- Simply regenerate the affected ensemble if you want to alter it (this is the intended method). The pure generation of a pulsed object is negligibly fast. I would strongly recommend this way because it is also the one to be guaranteed working in future qudi versions.

logic/pulsed/sequence_generator_logic.py
kay-jahnke: did a check, that the pulse_generator_settings are really the same.

logic/pulsed/sequence_generator_logic.py
kay-jahnke: deleted the `pulse_generator_settings` from the `ensemble_info`, as they are not in the `ensemble_info` for freshly sampled ensembles. The `written_waveforms` are left in, as they are added afterwards by hand anyways.

logic/pulsed/sequence_generator_logic.py
Neverhorst: Looks ok... I'm no big fan of those unreadable logic expressions though.

PR #517, opened 2019-05-28
Changed all channel config sets in pulser constraints to `frozenset` type

Neverhorst:
## Description
This resolves a bug when checking local and netref (rpyc) sets for equality.
Updated the interface documentation and dummy module as well.
Does not break in case of non-remote access if standard sets are used.

## Motivation and Context
Comparing a local set `a` with an rpyc netref set `b` by using `a == b` (or `b == a`) returns `False` even if the contents are the same (same hashes in both sets but iterator yields different ordering).
This happens despite the fact that `a.issubset(b) and a.issuperset(b)` or `b.issubset(a) and b.issuperset(a)` both return `True` (equality).
Since the only difference is the element order when iterating over both sets I assume this is a bug in the implementation of the equality operator in cpython. 
rpyc does not overwrite any equality checks but just creates a weakref to the corresponding methods.

## How Has This Been Tested?
Dummy Win8.1 x64
Experimental Setup Win8.1 x64 with Keysight AWG

## Types of changes

## Checklist:

### DIFF
--- a/hardware/awg/keysight_M3202A.py
+++ b/hardware/awg/keysight_M3202A.py
@@ -146,7 +146,7 @@ class M3202A(Base, PulserInterface):
         constraints.sequence_steps.default = 1
 
         activation_config = OrderedDict()
-        activation_config['all'] = {'a_ch1', 'a_ch2', 'a_ch3', 'a_ch4'}
+        activation_config['all'] = frozenset({'a_ch1', 'a_ch2', 'a_ch3', 'a_ch4'})
         constraints.activation_config = activation_config
         # FIXME: additional constraint really necessary?
         constraints.dac_resolution = {'min': 14, 'max': 14, 'step': 1, 'unit': 'bit'}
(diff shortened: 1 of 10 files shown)


Neverhorst:
Here is also a small standalone example of the issue.
Run the rpyc service (`rpyctest.py`) in one process and connect to it using the notebook (`rpyctest.ipynb`).


[rpyctest.zip](https://github.com/Ulm-IQO/qudi/files/3228682/rpyctest.zip)

Neverhorst:
Ok, apparently _rpyc_ is causing much more headache than I thought. Seems like the only proper solution would be to teach _rpyc_ to create local copies of python objects (pickle and unpickle) by default or to use a different package entirely.

As for this PR: I would still like to merge this change since regardless of the _rpyc_ issue, the predefined sets of possible channel combinations should not be changed at runtime and therefore a frozenset is more suitable.

PR #468, opened 2019-01-07
Pid cleaning

alrik-durand:
## Description
I've changed the process_interface and process_control_interface to respect Qudi's naming conventions.
I've updates the hardware module accordingly.
I've added the unit getters fot the pid_controller_interface.

## Motivation and Context
These interfaces are not used a lot yet. I've updated them before future development.

## How Has This Been Tested?
I've tested this on a new code using software_pid_controller (to come). I have not tested the modified hardware modules as I don't have the hardware to test it.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] I have documented my changes in the changelog (`documentation/changelog.md`)
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have added/updated for the module the config example in the docstring of the class accordingly.
- [x] I have checked that the change does not contain obvious errors (syntax, indentation, mutable default values).
- [ ] I have tested my changes using 'Load all modules' on the default dummy configuration with my changes included.
- [ ] All changed Jupyter notebooks have been stripped of their output cells.

PR #519, opened 2019-06-03
Power supply Keysight E3631A hardware file

alrik-durand:
## Description
I've added a hardware file to interface a power supply Keysight E3631A.

## Motivation and Context
This change add support to new instrument throught the Qudi ProcessControlInterface.

## How Has This Been Tested?
I've used this on the instrument with a Windows 10 PC.

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -23,7 +23,8 @@ Purely for displaying purposes; raw data is not affected by this filter.
 * analyze functions can be called either with the appropriate objects or with the object name
 * while sampling a sequence, the ensembles are only sampled if they weren't already sampled before
 * Add `natural_sort` utility function to `core.util.helpers`
-* 
+* Added a hardware file for power supply Keysight E3631A with a process control interface
+*
 
 Config changes:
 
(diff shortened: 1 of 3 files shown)


Neverhorst:
I see no problem in adding a new hardware module.

I have a request, though... Could you please change all method names (search and replace for all modules) related to the `ProcessControlInterface` and `ProcessInterface` to qudi naming convention?
It just occurred to me that these interfaces use _CamelCase_ convention but for qudi we decided all method names should be lower case and words separated by underscores.
This should be done as soon as possible before more people write modules based on these interfaces.

alrik-durand:
Thank you for you answer.
As you have mentionned [before ](https://github.com/Ulm-IQO/qudi/pull/468), a change in an interface should not be done slightly, so I kept the old naming convention for these PR. 
I agree this need to be done anyway, maybe I can do it right away in a separate PR ?

Neverhorst:
Yes, I said that. ;)
But in case of the `Process(Control)Interface` I would sneak this change in as early as possible because it is not (barely) being used so far.
A separate PR is also fine, thank you.

Neverhorst:
Could you please resolve the `changelog.md` conflict?

alrik-durand:
This should do it !

PR #522, opened 2019-06-06
Omniscan gui dev

Neverhorst:
## Description
This is a start to rework the confocal/scanning toolchain towards more general scanning measurements.
This PR introduces a new GUI module along with a very crude (for testing purposes only) logic module to provide a user interface for scanning with an arbitrary number of axes. Not all features are fully functional yet (optimizer, tilt correction etc.) but it's a start.
The next steps would be to think about a proper high-level logic (the one the GUI will connect to) and after that the GUI can be adapted accordingly. But the general GUI concept shouldn't change too much I think.

## How Has This Been Tested?
Win8.1 x64 dummy (default cfg)

## Screenshots:
Capture

## Types of changes

## Checklist:


logic/scanning_logic.py
```diff
@@ -0,0 +1,629 @@
...
+        with self.threadlock:
+            if len(self.__running_scan) > 1:
+                max_number_of_lines = self.scanner_settings['scan_resolution'][self.__running_scan[1]]
+            else:
+                max_number_of_lines = self.scanner_settings['scan_resolution'][self.__running_scan[0]]
+            if self.__scan_line_count >= max_number_of_lines or self.__scan_stop_requested:
+                while len(self._history) >= self.max_history_length:
+                    self._history.pop(0)
+                self._history.append(copy.deepcopy(self.scan_data[self.__running_scan]))
+                self._history_index = len(self._history) - 1
```
kay-jahnke: wouldn't that cause the history index to be always max_history_length, even if the history is not filled in the beginning?

logic/scanning_logic.py
```diff
@@ -0,0 +1,629 @@
...
+                self.module_state.unlock()
+                self.sigScanStateChanged.emit(False, self.__running_scan)
+                self.__timer.start()
+                return
+
+            if len(self.__running_scan) > 1:
+                y_min, y_max = self.scanner_settings['scan_range'][self.__running_scan[1]]
```
kay-jahnke: by working with self.__running_scan[-1] you can get around the whole business of distinguishing
`if len(self.__running_scan) > 1:`

logic/scanning_logic.py
```diff
@@ -0,0 +1,629 @@
...
+            result *= amplitude - offset
+            result += offset
+            return result
+
+        xx, yy = np.meshgrid(np.linspace(x_start, x_end, x_res),
+                             np.linspace(y_start, y_end, y_res),
+                             indexing='ij')
+        return np.random.rand(xx.shape[0], xx.shape[1]) * amplitude * 0.10 + gauss_ensemble(xx, yy)
+
+    def generate_1d_dummy_data(self, axis):
```
kay-jahnke: why underscore the generate function of the 2d data but not the 1d data?

logic/scanning_logic.py
```diff
@@ -0,0 +1,629 @@
...
+from core.module import Connector, ConfigOption, StatusVar
+
+
+class ScanData:
+    """
+
+    """
+    def __init__(self, scan_axes, channel_config, scanner_settings):
+        self.timestamp = datetime.datetime.now()
+        self._scan_axes = tuple(scan_axes)
```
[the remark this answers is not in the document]
Neverhorst: Since the ordering matters (which scan axis is displayed in x-/y-direction), no.

logic/scanning_logic.py
```diff
@@ -0,0 +1,629 @@
...
+        if self._scan_axes not in scanner_settings['scan_axes']:
+            raise ValueError('scan_axes must be tuple of axes name strings contained in '
+                             'scanner_settings')
+        self._target_ranges = tuple(scanner_settings['scan_range'][ax] for ax in self._scan_axes)
+        self._resolution = tuple(scanner_settings['scan_resolution'][ax] for ax in self._scan_axes)
+        self._channel_names = tuple(channel_config)
```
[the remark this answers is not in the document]
Neverhorst: I don't think default values are applicable here. These parameters are mandatory for a scan and so it's fine to throw an error if they are absent.

Regarding the use of tuples:
Since all these parameters should represent a snapshot during __init__ and must stay the same as long as this instance exists, I found it suitable to copy the values into a non-mutable data type. This would also avoid possible errors in the future if the dict values are mutable data types.
In addition, having a couple of instance attributes increases readability compared to a single dict in my opinion.

logic/scanning_logic.py
Neverhorst: I don't see why it should be. `self._history_index = len(self._history) - 1` is assigning the current length of the history list minus 1. So if there was no history entry before, append will add the first entry, then `len(self._history) == 1` and so `self._history_index == 0`.

logic/scanning_logic.py
Neverhorst: Ok, how?
I see how 
```
x_min, x_max = self.scanner_settings['scan_range'][self.__running_scan[-1]]
```
will work but the assignment of
```
self.__scan_line_positions[self.__running_scan[1]] = ...
```
is still different.

logic/scanning_logic.py
Neverhorst: Well, this logic module is just an intermediate solution to demonstrate the GUI functionality. This is not polished and never will be since it will be discarded in the near future. The only part that will probably be reused is the `ScanData` class.

Neverhorst:
Just to clarify:
The logic module introduced in this PR is purely for demonstration of the GUI functionality. It is not intended as basis for OMNISCAN. The only part that will probably be carried over in a new scanning logic implementation is class `ScanData`.

logic/scanning_logic.py
kay-jahnke: my bad. sorry, you are correct.

logic/scanning_logic.py
kay-jahnke: Ah damn, yes I did not pay attention to the fact that you are using `scan_resolution` in one case and `__scan_line_positions` in the other.

LukePiewalker:
We/ @konstantinfehler noticed that in this gui design (as far as I see it now it is the same as in https://github.com/Ulm-IQO/qudi/tree/core_pyside2_with_modules) till yet there is no more "resume scan" action. Is this omitted intentionally or should this feature still exist?

Neverhorst:
> We/ @konstantinfehler noticed that in this gui design (as far as I see it now it is the same as in https://github.com/Ulm-IQO/qudi/tree/core_pyside2_with_modules) till yet there is no more "resume scan" action. Is this omitted intentionally or should this feature still exist?

First of all, I think we should close this PR. It was meant as a proposal and kickoff for the work on OMNISCAN. Since it coincided with the rework of the qudi core (which requires a lot of rework on GUI modules), this PR does not make sense anymore.

You are right, @LukePiewalker, that the general concept of this GUI has been transferred from this PR to the branch [core_pyside2_with_modules](https://github.com/Ulm-IQO/qudi/tree/core_pyside2_with_modules "core_pyside2_with_modules"). However it has been developed and changed quite a lot since then.

It was not intended to exclude the "resume scan" functionality alltogether, but it was left out until a proper concept on a high-level logic is coming up. Since the new scanning logic is handling each scan as an encapsulated job, resuming it proves a bit more difficult than before. Also most non-direct scanning schemes (e.g. through third party API) can not handle resuming a scan, which needs to be handled separately.

To sum it up:  
It is intended to provide "resume scan" functionality, but I did not regard it as high-priority target for OMNISCAN.

Neverhorst:
Obsolete

PR #528, opened 2019-06-18
Non blocking jupyter output

kay-jahnke:
So far any notebook running in a jupyter kernel was waiting for the cell to finish before it printed any output. In this new version any output of a cell is now printed just after is has been generated.
In the process I also cleaned up the code by automatic code cleaning (Ctrl+Alt+L) and moved the communication relevant methods into a helper class.

## Description

Things still to do in a separate PR:
  * Under linux the notebook might hang when multiple cells are executed at the same time (for some reason under Windows it work. I suspect the QSocketNotifier dropping messages)
  * Implement support for magic commands
  * Implement matplotlib live plotting functionality (ion)

## Motivation and Context

Especially for cells that run long experiments in a while loop, it is very useful to see output live in order to follow the experiment and adjust if needed.

## How Has This Been Tested?

dummy, Linux 5.1.1-2-MANJARO
N25 Confocal, Win10 64x

## Types of changes

## Checklist:


logic/jupyterkernel/stream.py
```diff
@@ -78,3 +89,167 @@ def close(self):
...
+        return identities, m
+
+
+class IOStdoutNetworkStream:
+    _output_channel = 'stdout'
+
+    def __init__(self, network_stream):
+        self._network_stream = network_stream
+
+    def write(self, s):
```
drogenlied: This function needs to make sure that it does not call send from a thread that the socket does not belong to.

print() is not generally thread safe, it just works most of the time and does not crash, but there are no guarantees about doubled characters or lines, so it does not matter if our stdout replacement is thread safe.

The underlying socket is different, mixing up stream contents is a bad thing and sockets should only ever be read and written in a single thread.

logic/jupyterkernel/stream.py
```diff
@@ -78,3 +89,167 @@ def close(self):
...
+        content = {
+            'name': self._output_channel,
+            'text': s,
+        }
+        self._network_stream.send(msg_type='stream', content=content)
+
+    def flush(self):
```
drogenlied: An explanation why this can be empty would be nice.

logic/jupyterkernel/stream.py
```diff
@@ -78,3 +89,167 @@ def close(self):
...
+        pass
+
+    @property
+    def readable(self):
```
drogenlied: Are these properties general or situational?

Does readable mean that this is readable in general but may be not be ready now or does it mean that there are bytes that can be read?

logic/jupyterkernel/stream.py
```diff
@@ -78,3 +89,167 @@ def close(self):
...
+        return False
+
+    @property
+    def writable(self):
```
drogenlied: Same question as for readable.

Neverhorst:
I have tested it and everything is working. Except rapidly calling `print()` is causing glitches (parts of or entire prints lost). This may be related to the thread safety issue mentioned by @drogenlied.
You can reproduce that by using the attached notebook:

[kernel_testing.zip](https://github.com/Ulm-IQO/qudi/files/3354538/kernel_testing.zip)

I have conducted my tests on a Win8.1 x64 PC.

kay-jahnke:
Thank you very much. I will implement a bit more thread safety and then come back to you.

logic/jupyterkernel/stream.py
```diff
@@ -78,3 +89,167 @@ def close(self):
...
+        parts = [self.DELIM,
+                 signature,
+                 msg_lst[0],
+                 msg_lst[1],
+                 msg_lst[2],
+                 msg_lst[3]]
+        if identities:
+            parts = identities + parts
+        logging.debug('{0!s} send parts: {1!s}'.format(self.name, parts))
+        self.socket.send_multipart(parts)
```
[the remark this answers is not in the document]
kay-jahnke: This has been made threadsafe by wrapping it in Mutex.

logic/jupyterkernel/stream.py
kay-jahnke: Very good point. I fixed that, see below.

Neverhorst:
The last commit actually fixes the issue I mentioned.

kay-jahnke:
I added thread safety on two accounts:
1. The IO redirect was completely reworked. Now we inherit from StringIO and therefore have an internal buffer. On initialization a thread is created that regularly queries this buffer and calls the network function. The thread safety is guaranteed by using locks in the same class. This also reduces the number of transmitted messages to a reasonable amount.
2. The actual send command for the network traffic was wrapped into a Mutex statement, to guarantee it not being called twice at the same time. This was necessary because although the IOStream itself might be thread safe, both `stderr` and `stdout` access the same stream.

Nasty subtleties: During the execution of code inside (ICode) a cell the Qt-Thread is logged in (at least on my linux machine). This means, if you fire a QtSignal to a class outside this ICode, it will only be executed after the cell has finished. This is also the reason why inside the cell you can use `threading` without it colliding with `QThread`.

Testing result: The redirected output is very fast now. If you run prints in a loop without any sleep, you can produce so much output, that you hit the transmission limit of ZQM and it complains. This I would call intended behavior.

@Neverhorst : I tried your example and for me it now runs flawless under linux.
@drogenlied : Can you please check again.

PR #530, opened 2019-06-25
Integration of MicrowaveInterface into AWG70K module

Neverhorst:
## Description
Integrated the `MicrowaveInterface` into the AWG70k hardware module in order to perform CW ODMR measurements using only the AWG.

Changed the `MicrowaveInterface` method `get_status` to something more distinguishable (`get_microwave_status`) in order to avoid namespace collisions when combining `PulserInterface` and `MicrowaveInterface` into a single module. Changed all occurrences of this method name.

Also changed `OdmrLogic` to start the timer AFTER all devices are active and running. Loading and starting a sequence with >100 waveforms can take some time with the AWG and so the timer has to wait for this process to finish in order to represent actual measurement time.

**WARNING:** 
This PR is a precedent case for namespace collisions when using several interfaces in a single module.
Before merging this PR and messing up peoples microwave scripts and custom hardware modules, we should discuss how we want to proceed with this issue. Maybe we should directly reform qudi in order to fix the current situation. This issue will otherwise persist and cause a steady tickle of interface changes in the future. This behavior contradicts the qudi ideal. This is also why I did not update the documentation yet.
So please discuss...

## Motivation and Context
I went for a hardware implementation rather than an interfuse using `SequenceGeneratorLogic` or `PulsedMasterLogic` because the implementation is dependent on some features unique to the _Tektronix AWG70k_ series that are not available through the `PulserInterface`.
It utilizes the device specific "basic waveform compiler" plug-in for faster creation of a large number of short and simple sine waveforms as well as the software triggering capability of sequences.

## How Has This Been Tested?
IQO Setup 9, Win8.1 x64

## Types of changes

## Checklist:

### DIFF
--- a/hardware/awg/tektronix_awg70k.py
+++ b/hardware/awg/tektronix_awg70k.py
@@ -1894,8 +1894,7 @@ class AWG70K(Base, PulserInterface, MicrowaveInterface):
 
         self.__mw_mode = MicrowaveMode.LIST
 
-        if frequency is not None and any(
-                freq != frequency[ii] for ii, freq in enumerate(self.__mw_list_params['freq'])):
+        if frequency is not None and tuple(frequency) != tuple(self.__mw_list_params['freq']):
             self.__mw_list_params['freq'] = frequency
             self.__mw_list_params['generated'] = False
         if power is not None and power != self.__mw_list_params['power']:
--- a/hardware/awg/tektronix_awg70k.py
+++ b/hardware/awg/tektronix_awg70k.py
@@ -2195,6 +2195,11 @@ class AWG70K(Base, PulserInterface, MicrowaveInterface):
             samples_per_period = sample_rate / freq
             cycles = max(100, int(np.ceil(2 * self.__min_waveform_length / samples_per_period)))
 
+            samples = np.round(cycles * samples_per_period)
+
+            while samples % 2 != 0:
+                cycles += 1
+                samples = np.round(cycles * samples_per_period)
             # Write waveform for channel 1 and add to sequence
             name = '_mw_sweep_{0:d}_ch1'.format(ii)
             if a_ch == 1:
--- a/hardware/awg/tektronix_awg70k.py
+++ b/hardware/awg/tektronix_awg70k.py
@@ -22,6 +22,7 @@ top-level directory of this distribution and at <https://github.com/Ulm-IQO/qudi
 
 
 import os
+import re
 import time
 import visa
 import numpy as np
@@ -34,9 +35,10 @@ from core.module import Base, ConfigOption
 from core.util.modules import get_home_dir
 from core.util.helpers import natural_sort
 from interface.pulser_interface import PulserInterface, PulserConstraints
+from interface.microwave_interface import MicrowaveInterface, MicrowaveLimits, MicrowaveMode, TriggerEdge
 
 
-class AWG70K(Base, PulserInterface):
+class AWG70K(Base, PulserInterface, MicrowaveInterface):
     """ A hardware module for the Tektronix AWG70000 series for generating
         waveforms and sequences thereof.
 
@@ -67,6 +69,10 @@ class AWG70K(Base, PulserInterface):
     _username = ConfigOption(name='ftp_login', default='anonymous', missing='warn')
     _password = ConfigOption(name='ftp_passwd', default='anonymous@', missing='warn')
 
+    # cw microwave config options
+    _microwave_trigger = ConfigOption(name='microwave_trigger', default='A', missing='warn')
+    _laser_channel = ConfigOption(name='laser_channel', default='d_ch1', missing='warn')
+    _microwave_channel = ConfigOption(name='microwave_channel', default='a_ch1', missing='warn')
 
     # translation dict from qudi trigger descriptor to device command
     __event_triggers = {'OFF': 'OFF', 'A': 'ATR', 'B': 'BTR', 'INT': 'INT'}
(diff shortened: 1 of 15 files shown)


hardware/awg/tektronix_awg70k.py
```diff
@@ -1678,3 +1723,487 @@ def _create_xml_header(self, number_of_samples, markers_active):
...
+            self.query('*OPC?')
+            set_freq = float(self.query('BWAV:FREQ?'))
+            set_cycles = int(float(self.query('BWAV:CYCL?')))
+
+        # Compile waveform and set laser channel marker to high
+        self.write('BWAV:COMP')
+        self.write('*WAI')
+        self.query('*OPC?')
+        a_ch = int(name.rsplit('_ch', 1)[-1])
+        laser_ch = int(self._laser_channel.rsplit('ch', 1)[-1])
```
kay-jahnke: What happens if the laser channel is empty?

hardware/awg/tektronix_awg70k.py
```diff
@@ -1678,3 +1723,487 @@ def _create_xml_header(self, number_of_samples, markers_active):
...
+                raise Exception('Microwave channel must be either "a_ch1" or "a_ch2".')
+            if a_ch == 1:
+                self._compile_sine_waveform('_mw_cw_ch1', freq, amp, cycles)
+                self._compile_sine_waveform('_mw_cw_ch2', freq, 0, cycles)
+            else:
+                self._compile_sine_waveform('_mw_cw_ch1', freq, 0, cycles)
+                self._compile_sine_waveform('_mw_cw_ch2', freq, amp, cycles)
+        else:
+            self._compile_sine_waveform('_mw_cw_ch1', freq, amp, cycles)
+        print(self.awg.query('SYSTem:ERRor:ALL?'))
```
kay-jahnke: print left over from debugging?

hardware/awg/tektronix_awg70k.py
```diff
@@ -1678,3 +1723,487 @@ def _create_xml_header(self, number_of_samples, markers_active):
...
+
+        self._init_waveform_compiler()
+
+        amp = self.dbm_to_volts(power)
+        a_ch = int(self._microwave_channel.rsplit('_ch', 1)[-1])
+        sample_rate = self.get_sample_rate()
+
+        self.new_sequence('_mw_sweep', len(frequencies) + 1)
+        for ii, freq in enumerate(frequencies, 1):
+            print(freq)
```
kay-jahnke: debug print?

hardware/awg/tektronix_awg70k.py
```diff
@@ -1678,3 +1723,487 @@ def _create_xml_header(self, number_of_samples, markers_active):
...
+            # Calculate waveform length (as number of periods) for this frequency step
+            samples_per_period = sample_rate / freq
+            cycles = max(100, int(np.ceil(self.__min_waveform_length / samples_per_period)))
+
+            # Write waveform for channel 1 and add to sequence
+            name = '_mw_sweep_{0:d}_ch1'.format(ii)
+            if a_ch == 1:
+                print(self._compile_sine_waveform(name, freq, amp, cycles))
+            else:
+                print(self._compile_sine_waveform(name, freq, 0, cycles))
```
kay-jahnke: more debug prints?

hardware/awg/tektronix_awg70k.py
```diff
@@ -116,6 +128,18 @@ def on_activate(self):
         self.__max_seq_repetitions = int(self.query('SLIS:SEQ:STEP:RCO:MAX?'))
         self.__min_waveform_length = int(self.query('WLIS:WAV:LMIN?'))
         self.__max_waveform_length = int(self.query('WLIS:WAV:LMAX?'))
+        self.__has_sequence_mode = '03' in self.query('*OPT?').split(',')
+
+        # CW microwave attributes
+        self.__mw_mode = MicrowaveMode.CW
+        self.__mw_cw_params = {'freq': 2.87e9, 'power': -30}
+        self.__mw_sweep_params = {'start': 2.82e9, 'stop': 2.92e9, 'step': 2e6, 'power': -30}
+        self.__mw_list_params = {'freq': list(np.linspace(2.82e9, 2.92e9, 101)), 'power': -30}
```
kay-jahnke: Why set these parameters in on_activate and __init__ twice?

hardware/awg/tektronix_awg70k.py
Neverhorst: removed

hardware/awg/tektronix_awg70k.py
Neverhorst: removed

hardware/awg/tektronix_awg70k.py
Neverhorst: removed

hardware/awg/tektronix_awg70k.py
Neverhorst: I just made it a habit to declare all instance attributes in `__init__` and initialize them first as default/empty data type (empty list/dict or `None` for more complex types). The "proper" initialization of these attributes follows in `on_activate`. 
But this way you can see all available variables easily at once. 
Additionally it satisfies code inspection tools since you should avoid declaring instance variables outside of `__init__`.
I could stick to my usual pattern and just initialize them as empty dicts in `__init__`.

hardware/awg/tektronix_awg70k.py
Neverhorst: Well first of all it's a protected variable so any changes you do to it from outside the module are generally at your own risk.

I added some sanity checks to the activation to avoid corrupted AppStatus.

Neverhorst:
Closed until PR #542 has been merged into master. I will adapt this PR afterwards to the newly introduced features.

PR #533, opened 2019-07-01
Made natural_sort helper function more robust

Neverhorst:
## Description
Made `natural_sort` helper function more robust by type checking for non-str type and falling back to plain `sorted()` in that case.

## Motivation and Context
`natural_sort` failed when handing over an array of scalars.

## How Has This Been Tested?
On dummy

## Types of changes

## Checklist:

### DIFF
--- a/core/util/helpers.py
+++ b/core/util/helpers.py
@@ -210,6 +210,10 @@ def natural_sort(iterable):
     @param str[] iterable: Iterable with str items to sort
     @return list: sorted list of strings
     """
+    # If first element of iterable is not a string, fall back to simple sorted()
+    if len(iterable) == 0 or not isinstance(iterable[0], str):
+        return sorted(iterable)
+
     def conv(s):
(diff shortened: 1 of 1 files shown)


alrik-durand:
This piece of code generates an error when I load POI GUI. I think if iterable is a dict (in the case of POI GUi for example, then len(iterable) is 1 but iterable[0] generates a KeyError.
For this module this can be fixed with :

```
    if len(iterable) == 0 or (not isinstance(iterable, dict) and not isinstance(iterable[0], str)):
        return sorted(iterable)
```
But I'm not sure that this will solve all the possible cases !

alrik-durand:
Pulse measurements also fails to load, 
sequence_generator_logic >> _update_blocks_from_file >> natural_sort
iterable is then a generator, which "has no len()" says the error message :(

Neverhorst:
Man, you are so right. That was a prime example of rushed changes... Just opened up PR #535 to revert changes.

PR #534, opened 2019-07-02
Adding oceanoptics spectrometer hardware file

jochen-scheuer:
## Description

Adding a hardware file for the ocean optics spectrometer

## Motivation and Context

New hardware.

## How Has This Been Tested?

Tested by using the spectrometer QE-Pro on Windows10 with two different computers.

## Types of changes

## Checklist:


hardware/spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,85 @@
...
+        module.Class: 'spectrometer.oceanoptics_spectrometer.OceanOptics'
+        spectrometer_serial: 'QEP01583' #insert here the right serial number.
+
+    """
+    _serial = ConfigOption('spectrometer_serial', missing='warn')
+
+    def on_activate(self):
+        """ Activate module.
+        """
+        self.integration_time = 1e4
```
Neverhorst: Hard-coding measurement parameters (especially in hardware modules) is problematic.
Isn't this initialization of the integration time unnecessary anyways? The device has a time set regardless of qudi.

hardware/spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,85 @@
...
+        return specdata
+
+    def getExposure(self):
+        """ Get exposure.
+
+            @return float: exposure
+
+            Not implemented.
+        """
+        return self.integration_time_micros
```
Neverhorst: Right after `__init__` this module instance has no attribute `integration_time_micros` and thus will raise `AttributeError`.
Always declare all instance variables in `__init__` first to avoid such a thing.

But I don't see why this variable is necessary? `integration_time_micros` seems to be stored in `spec` as well. So why not simply set and get this value in/from `spec` and avoid an additional copy which can possibly deviate from the true value set in the device?

hardware/spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,85 @@
...
+
+from core.module import Base, ConfigOption
+from interface.spectrometer_interface import SpectrometerInterface
+import numpy as np
+import seabreeze.spectrometers as sb
+
+
+
+class OceanOptics(Base, SpectrometerInterface):
+    """ Hardware module for reading spectra from the WinSpec32 spectrometer software.
```
Neverhorst: I'm not sure but does the _Ocean Optics_ spectrometer use _WinSpec32_ or is this just a copy-paste leftover from another module?

hardware/spectrometer/oceanoptics_spectrometer.py
```diff
@@ -0,0 +1,85 @@
...
+
+Qudi is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with Qudi. If not, see <http://www.gnu.org/licenses/>.
+
+Derived from the pyPL project (https://github.com/kaseyrussell/pyPL)
```
Neverhorst: Is this really related to this module?

kay-jahnke:
Can you please also add a changelog and perform the tests defined by the checkbox list.

hardware/spectrometer/oceanoptics_spectrometer.py
jochen-scheuer: Actually, this is an important parameter and one has to be able to change it to work with the spectrometer. It should be possible to change this value in the gui. I changed it into a status variable such that it at least does not have to be set everytime.

hardware/spectrometer/oceanoptics_spectrometer.py
jochen-scheuer: Was an error and is fixed, Thanks.

hardware/spectrometer/oceanoptics_spectrometer.py
jochen-scheuer: Also a copy paste error, got fixed.

hardware/spectrometer/oceanoptics_spectrometer.py
jochen-scheuer: fixed with StatusVar

hardware/spectrometer/oceanoptics_spectrometer.py
Neverhorst: I was not suggesting to get rid of the possibility to change the exposure time.
I was just asking if you can read back this value from `spec`. If you can set the value with `self.spec.integration_time_micros(my_value)` there may also be a possibility to read `self.spec.integration_time_micros`. This would get rid of a possible error source and reduce redundancy by getting rid of the instance variable `self.integration_time`.
You would then still be able to set and get the integration time by calling `setExposure` and `getExposure`.

kay-jahnke: I looked into the code of the Seabreeze module: https://github.com/ap--/python-seabreeze/blob/master/seabreeze/spectrometers.py
Interestingly it does not have a way to query the integration time. Therefore I think Jochens implementation is valid.

Neverhorst: Ok, cool, that's all I wanted to know. In that case the current solution is the best option.

PR #535, opened 2019-07-03
Revert "Made natural_sort helper function more robust"

Neverhorst:
Reverts Ulm-IQO/qudi#533 due to hastily introduced errors

### DIFF
--- a/core/util/helpers.py
+++ b/core/util/helpers.py
@@ -212,4 +212,7 @@ def natural_sort(iterable):
     """
     def conv(s):
         return int(s) if s.isdigit() else s
-    return sorted(iterable, key=lambda key: [conv(i) for i in re.split(r'(\d+)', key)])
+    try:
+        return sorted(iterable, key=lambda key: [conv(i) for i in re.split(r'(\d+)', key)])
+    except:
+        return sorted(iterable)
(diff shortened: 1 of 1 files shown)


kay-jahnke:
How about the upgraded fix?

I agree, the usage of the natural_sort on non-string objects is the problem, but I would like the helper to be more robust and actually somehow cope with that.

Neverhorst:
Shame! _\*dingdingding*_ Shame! _\*dingding*_ Shame!

I agree with @drogenlied. We should never do such a search and replace rampage again without thorough testing.

However I just thought about the same thing @kay-jahnke just committed. I think this is the best solution. (Would be even better to use explicit exceptions)

kay-jahnke:
@alastair-marshall and @alrik-durand :
Could you please test this case for your  setup?

alrik-durand:
No more problem here !

Neverhorst:
I think the fix by @kay-jahnke is pretty much bulletproof since it truly falls back to `sorted()` upon a somehow failed natural sorting. This should basically undo all the damage done by replacing all occurrences of `sorted` with `natural_sort`.

Can we all agree on merging this?

alrik-durand:
Yes please !

PR #536, opened 2019-07-08
Extension to pulsed module

kay-jahnke:
This PR contains several small extensions to the pulsed modules. The goals was to make the tool chain more flexible. In the following the changes will be described.

### Multiple predefined paths

The config options of the SequenceGeneratorLogic `additional_predefined_methods_path` and `additional_sampling_functions_path` now accept either a path as string or multiple paths as a list of strings.
#### Motivation and Context

There are two config options (`additional_predefined_methods_path` and `additional_sampling_functions_path`) to load additional function from folders outside of the qudi folder. Formerly they expected a string with the path to additional functions. Then qudi did load all the functions in that folder. There was no way to cherry pick between different sets of functions. This is relevant for example, if you need a second analogue channel on some hardware but not on others, but still want to use the same repository.
Now you can either specify the path as before or you specify a list of paths. Example:
``` yaml
additional_sampling_functions_path:
   - 'path/to/local/functions'
   - 'path/to/functions/in/custom/git'
```

### Use helper functions in predefined
qudi delivers some standard predefined functions. They are now making use of the helper function `_add_trigger`.
#### Motivation and Context
The predefined functions were written at an earlier point in time, when the helpers did not exist. By using the helper, first we save code and if a bug or change appears we only need to fix it at one place now.
Also, if you want to have an custom trigger for all your sequences, you can now simply overwrite the `_add_trigger` function and all predefined methods will use that custom trigger.

### New extraction and analysis methods
New extraction and analysis methods were added to the defaults delivered with qudi.
#### Extraction
- The existing method `gated_conv_deriv` got a parameter flank width. This widens (or narrows for negative values) the detected pulse. Currently, as the flank detection is done on the highest gradient, the laser pulse starts at half the signal height. There are some cases where it is beneficial to include the whole pulse.
- The method `ungated_pass_through` passes through any data coming from the hardware. This function is useful when the extraction and analysis are done in hardware but the results should still be handled by qudi.
- The method `gated_pass_through` passes through the data just as `ungated_pass_through` now for the gated case.
#### Analysis
- The method `analyse_pass_through` just passes through the data received from extraction. 2D data is summed along the second axis, while 1D data is just handed on. This is also for the case, where the analysis is done in hardware.
- The method `analyse_mean_reference` is essentially an mean of the selected signal window. Instead of dividing by the reference window to find the norm, the background window is subtracted to find the referenced mean.

### Improving pulsed load robustness
Formerly, when the SequenceGeneratorLogic had an error on activation, also a second error was thrown by the PulsedMasterLogic. This was because the master could not find some properties in the not fully loaded module. By explicitly using the `getattr()` function on these properties and returning an empty `dict` by default, we can at least make the PulsedMasterLogic  load correctly without throwing errors.

### Bug fixes
- The sampling function `sin` now accepts an arbitrary phase.
- Fixed the magnet gui to align with the changes done to the plot widget earlier.
- The pulser gui and also the fits through errors when invoke settings was activated and the loaded control variables had a length of 0. Fixed that by explicitly catching that case.
- When loading fits from file was aborted, that caused errors in the pulser, so I caught that case.
- Better error handling for missing dependencies in pickled files.

## How Has This Been Tested?

All changes have been tested on Linux 5.1.8-1-MANJARO with dummy config (although modified to test the new config options).
The changes are also in productive use on the mesoscopic setup: Win10 64x

## Types of changes

## Checklist:


kay-jahnke:
Deleted the `force_sequence_mode` option from this branch and opened a new branch for that.

Please now go ahead with your review.

logic/pulsed/pulsed_measurement_logic.py
```diff
@@ -991,6 +991,10 @@ def do_fit(self, fit_method, use_alternative_data=False, data=None):
         else:
             update_fit_data = False
 
+        if len(data) < 2 or len(data[0]) < 2 or len(data[1]) < 2:
+            self.debug.error('The data you are trying to fit does not contain enough data for a fit.')
```
Neverhorst: What shall it be? An error or a debug message? In any case `debug` is not an attribute of `PulsedMeasurementLogic` and so this line leads to an `AttributeError`.
I would suggest `self.log.debug(...)`

logic/pulsed/sequence_generator_logic.py
```diff
@@ -117,10 +118,19 @@ def __init__(self, config, **kwargs):
         # directory for additional generate methods to import
         # (other than logic.predefined_generate_methods)
         if 'additional_methods_dir' in config.keys():
-            if not os.path.exists(config['additional_methods_dir']):
+            if isinstance(config['additional_methods_dir'], str) and not os.path.exists(config['additional_methods_dir']):
```
Neverhorst: Instead of defining the same operation for the single path and the all paths in a for loop, I would simply wrap a single str path into a list of length 1 and always iterate over that.

logic/pulsed/sequence_generator_logic.py
```diff
@@ -157,7 +167,18 @@ def on_activate(self):
         # sampling functions from.
         sf_path_list = [os.path.join(get_main_dir(), 'logic', 'pulsed', 'sampling_function_defs')]
         if isinstance(self._sampling_functions_import_path, str):
-            sf_path_list.append(self._sampling_functions_import_path)
+            if not os.path.exists(self._sampling_functions_import_path):
```
Neverhorst: see above

logic/fit_logic.py
```diff
@@ -236,6 +236,8 @@ def load_fits(self, filename):
             @return dict: validated fit dictionary with function references and parameter objects
         """
         user_fits = OrderedDict()
+        if filename is None or filename == '':
```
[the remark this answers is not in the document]
kay-jahnke: jup, thats cleaner

logic/pulsed/pulsed_measurement_logic.py
kay-jahnke: must have been asleep in front of the keyboard, thanks

logic/pulsed/sequence_generator_logic.py
kay-jahnke: good point, I fixed it.

logic/pulsed/sequence_generator_logic.py
kay-jahnke: see above

logic/pulsed/sequence_generator_logic.py
```diff
@@ -117,10 +118,20 @@ def __init__(self, config, **kwargs):
         # directory for additional generate methods to import
         # (other than logic.predefined_generate_methods)
         if 'additional_methods_dir' in config.keys():
-            if not os.path.exists(config['additional_methods_dir']):
-                self.log.error('Specified path "{0}" for import of additional generate methods '
-                               'does not exist.'.format(config['additional_methods_dir']))
-                self.additional_methods_dir = None
+            if isinstance(config['additional_methods_dir'], str):
```
Neverhorst: I think this should probably move to `on_activate`. If you do so, please also check for the default value `None`.

And by the way, I just realized this is an old bug.
`'additional_methods_dir'` is never in `config.keys()` because the name given to the `ConfigOption` is actually `'additional_predefined_methods_path'`

logic/pulsed/sequence_generator_logic.py
```diff
@@ -157,7 +168,18 @@ def on_activate(self):
         # sampling functions from.
         sf_path_list = [os.path.join(get_main_dir(), 'logic', 'pulsed', 'sampling_function_defs')]
         if isinstance(self._sampling_functions_import_path, str):
-            sf_path_list.append(self._sampling_functions_import_path)
+            self._sampling_functions_import_path = [self._sampling_functions_import_path]
```
Neverhorst: Similar here. Please check for default value (`None`)

logic/pulsed/sequence_generator_logic.py
kay-jahnke: Good point, I completely overlooked that as I only tested with the ConfigOption.
Also followed your suggestion and moved the stuff to the on_activation and cleaned up a bit more.

logic/pulsed/sequence_generator_logic.py
kay-jahnke: see above.

PR #539, opened 2019-07-19
Fix prints in notebooks from threads

kay-jahnke:
The redirect of the output generated by the cells of a jupyther is now global for the kernel and not local to the cell. The redirect therefore also catches any output generated by threads started in a notebook cell, even after the cell has long finished.

## Description

The redirect is set up at the start of the kernel and only cleaned up together with the kernel at deactivation. By default **all** output is redirected to a StringIO which then checks, if the output is coming from inside the notebook or from somewhere else. The notebook output is send back to the browser, while the rest is send on to the normal stdout (same for stderr).
To recognize the output from a notebook, the imported Thread class was extended by a marker, so that any Thread created inside a notebook will have this marker. This marker is than only added to the notebook scripts.

## Motivation and Context

Creative people (me included) started using the jupyter notebooks for everything and created Threads inside notebooks. The prints of these could go missing, if the cell was already finished with the started thread still running (sometimes the output of these threads appeared in stdout instead of the notebook).

## How Has This Been Tested?

running test notebooks with threads and fast prints on dummy config: Linux 5.1.8-1-MANJARO

## Types of changes

## Checklist:

### DIFF
--- a/logic/jupyterkernel/helpers.py
+++ b/logic/jupyterkernel/helpers.py
@@ -29,6 +29,11 @@ distributed as part of this software.
 import sys
 import builtins
 from base64 import encodebytes
+import threading
+
+
+class ThreadFixer(threading.Thread):
+    notebook_thread = True
 
 
 class ExecutionResult:
--- a/logic/jupyterkernel/qzmqkernel.py
+++ b/logic/jupyterkernel/qzmqkernel.py
@@ -53,6 +53,7 @@ from warnings import warn
 import ast
 import traceback
 import jedi
+import threading
 
 # zmq specific imports:
 import zmq
@@ -185,6 +186,12 @@ class QZMQKernel(QtCore.QObject):
         self.displayhook = DisplayHook()
         self.display_trap = DisplayTrap(self.displayhook)
         self.builtin_trap = BuiltinTrap()
+        threading.current_thread().notebook_thread = True
+        self.stderr = RedirectedStdErr()
+        self.stderr.open(IOStderrNetworkStream(self.iopub_stream, sys.stderr))
+        self.stdout = RedirectedStdOut()
+        self.stdout.open(IOStdoutNetworkStream(self.iopub_stream, sys.stdout))
+
         setup_matplotlib(self)
 
     @QtCore.Slot()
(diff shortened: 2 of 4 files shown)


alrik-durand:
I think this PR induced a little problem : errors are sent to other cells, sometimes no cell at all.
For example, I define a function in one notebook
```
def do_something_bad():
    error()
```
Then I call in in another notebook
```
error()
```
The error is either sent to the output of the first notebook, either lost :(

kay-jahnke:
thanks @alrik-durand 

I had not tested that:
What I can reproduce is that, if I run one notebook it shows the output in the second. Actually it always shows it in the last notebook. I can also make the most current notebook hang (a star at the current number of executions) and than the output of any notebook is lost. What I can't reproduce is that the output is just lost somehow.

**Concerning further steps**: 
The whole problem relates to the fact that python has a single interpreter lock and therefore only knows a single stdout (or stderr for that matter). It does not differentiate between different sources of the output. Even the official IPython kernel has that problem.
There are however ways around that: You can for example use a proper logger (as qudi is doing in the background) and let it figure out the streams. This however does not work easily for just calling print. You could also rewrite the threading module (I did something quick and dirty in that direction for the current kernel) to tag and figure out where which stream goes. Probably there are even more solutions but whatever they are they are very complicated and by far not the focus of qudi. 
Therefore I'm not willing to invest large amounts of time to essentially reprogram a better IPython kernel inside qudi. This would need be done in its own project. 
There also is the idea to include the standard IPython shell into qudi and work with that if someone finds the time to do it. This is in my opinion the far better idea, as it supports magic commands and offers other advantages (not the least being, that we do not need to support it).

That said, I will give the problem a look and a bit of a poke and will see if I can come up with a better quick solution for qudi.

kay-jahnke:
@alrik-durand I checked simple solutions for the print in multiple notebooks again. The result is the branch https://github.com/Ulm-IQO/qudi/tree/fix_jupyter_multiple_kernels.
As it uses the hierarchical structure of the notebooks, this only works if you do not restart notebooks. Therefore I don't want to put it in master. To solve this for good, we would need to change the qudi core and give it a proper handling of all streams.

As this is a big project, I will defer it for now but open an issue to keep track of it.

PR #541, opened 2019-08-01
Add magnetic field calculation tool (cal_alignment) to ODMR module.

liuyanulm:
## Description

Added magnetic field calculation function in odmrlogic basing on the two dip values from a two-dip fiting, a Displaywidget in ODMR GUI to show the calculated magnetic field strength and the angle between N-V axis and external field.

## Motivation and Context

Our current ODMR measurement would not give an field result.
However, it is inconvenient for field estimation during some work. One has to do the calculation with external methods. Secondly, when we apply bias field to NV for routine NMR (for example, depth measurement), it would be annoying if we do not know the angle between N-V axis and the applied field.

## How Has This Been Tested?

Tested on dummy.
field cal

## Types of changes

## Checklist:


kay-jahnke:
@liuyanulm 
I like your idea, but I would actually go a step further and make it a full fledged functionality:
1. make the fit area a tabbed view
2. add a second tab for the field
3. give the calculation the option to calculate the values from given values
4. also include the calculation from field to line position
5. handle cases when you fit more or less than two lines (ensembles, hyperfine or high magnetic fields)
6. calculate errors on your values

Apart from these comments, we should think about what parameters we want to put to the options in order to not only limit ourselves to NV centers but support color centers in general (splitting, ...).

liuyanulm:
@kay-jahnke 
Thank you for the comment.
Your suggestions are inspiring...
Accordingly to your points 1,2,3, there are tabbed views for field calculation with ODMR freqs from fitted and from user input.
point 4 is bit of excessive...for experiments, not really practical...No offense..
Point 5, for other cases(less or more than 2 dip values), it would be more sophisticated, and new algorithm for analyzing the measured data may also needed. Therefore, it would go way beyond the initial proposal.
Point 6, For very precise field calculation, one has go to pulsed measurement. The initial idea of this PR is to assist people align bias field for NMR (the field precision and angle precision are very good at high field, since uncertainty caused by strain reduces relatively). As we all know the zero-field-splitting shifts with temperature (many case you do not even know how much heat your laser generates..), plus there is a strain factor, an accurate Dgs measurement actually is sophisticated... And these uncertainty convolute with dip-value uncertainty. Considering the amount work, and its functionality, I would really really like to pass this error calculation.

christoph-mueller:
Hey Yan,
I am quite sure that the way it is currently implemented it will only work for fields lower than around 1000G..., since the formula expects negativ frequencies for higher fields. The function's docstring even mentions this.

liuyanulm:
> Hi guys,
> 
> actually ODMR is a tool that is nearing its end of life as far as I can hear from many people. Most people use their own customized data visualization tools or directly use the Pulsed Measurement toolset since it is more versatile (pulsed, chirped, etc. ODMR). Apart from that qudi is not just for NV experiments and should remain as general as possible.
> 
> Calculating the magnetic field from a fit of +1 and -1 lines is definetely a useful feature but why make it so invasive (new GUI element just useful for NV experiments)?
> Why not simply make the field calculation a new fit function that has as free parameters the angle and field strength? I see only advantages in this approach:
> 
> 1. The user can decide if he wants to include this feature or not
> 2. The feature can be used in all modules that implement fitting (also for pulsed, chirped etc. ODMR)
> 3. It keeps the GUI clean

That's very thoughtful. I will do this.

liuyanulm:
> > Hi guys,
> > actually ODMR is a tool that is nearing its end of life as far as I can hear from many people. Most people use their own customized data visualization tools or directly use the Pulsed Measurement toolset since it is more versatile (pulsed, chirped, etc. ODMR). Apart from that qudi is not just for NV experiments and should remain as general as possible.
> > Calculating the magnetic field from a fit of +1 and -1 lines is definetely a useful feature but why make it so invasive (new GUI element just useful for NV experiments)?
> > Why not simply make the field calculation a new fit function that has as free parameters the angle and field strength? I see only advantages in this approach:
> > 
> > 1. The user can decide if he wants to include this feature or not
> > 2. The feature can be used in all modules that implement fitting (also for pulsed, chirped etc. ODMR)
> > 3. It keeps the GUI clean
> 
> That's very thoughtful. I will do this.

@Neverhorst Changing of mind after some second thought and tryouts.
(Your arguments did hit me, cuz this PR was quite some time ago. I nearly forgot the original motivations..)
First of all, the ODMR is mainly for NV in my mind. what else samples can run with ODMR?
For most ODMR calculation, the D and E would vary more or less from sample to sample. When you wish to have a perfect N-V axis alignment, these deviations of D and E do matter. However, my understanding for the current fitting setting does not allow customized parameter input, it come always with a left click of the mouse on "do fit" button.
Unfortunately, to make it handy, the input of individual D and E in a newly designed GUI would be very much helpful. Pulsed-ODMR by pulsedMeasurement, the user has to fill the +1 and -1 in the CW ODMR GUI for now. But I could not think better ideas for now. Any suggestion would be appreciated. (I even thought about creating an isolated calculation gui but perhaps too dramatic that way.)
If you try out this tool, it is really handy. 
Sincerely,
Yan

Neverhorst:
Well, the fact that fitting is pretty messed up is not a proper argument for new features that circumvent it. 
Fits in Qudi must be able to accept user inputs (GUI is already implementing that in fit setting dialogues). So I think the best way to proceed would be cleaning up how qudi implements fitting.

If you want a quick way of using this feature, implementing a separate GUI and Logic module would be the proper "Qudi way" to go. The overhead is minimal and it keeps the code clean and modular.

PR #542, opened 2019-08-12
Interface method decorator and core cleanup

Neverhorst:
## Description
This PR introduces a new concept to overload interface methods.
See new documentation _how_to_hardware_with_multiple_interfaces.md_ for more information on how to use this new feature.
There are some dummy modules for testing present which will be deleted before merging this PR into master:

- config/test.cfg
- hardware/myhardware.py
- interface/first_test_interface.py
- interface/second_test_interface.py

Also core folder/module structure has been organized a bit more in order to avoid circular dependencies and for consistency.

All this needs further testing since it impacts almost all modules.

**EDIT:** I also snuck in support for python native `frozenset` in the _yaml_ dumper/loader in order to support `frozenset` type StatusVars.

## Motivation and Context
When deriving a hardware module from multiple interfaces it is quite likely to run into problems with namespace collisions between these interface classes.
This PR enables you to define multiple implementations in a hardware module for a single interface method (occurring in multiple interfaces the modules inherits). The `Connector` object will hide this overload from logic modules, so nothing needs to be changed in those.

Many core classes have been moved to their own module and core paths have changed. Imports have been adjusted accordingly.

## How Has This Been Tested?
On Win8.1 x64 dummy and setup9
But needs more testing...

## Types of changes

## Checklist:


core/interface.py
```diff
@@ -64,11 +64,17 @@ def __get__(self, obj=None, cls=None):
         return self
 
     def __call__(self, *args, **kwargs):
+        if self.__isabstractmethod__:
+            raise NotImplementedError('No methods registered on abstractmethod {0}.'
+                                      ''.format(self.default_callable.__name__))
+        elif self.registered:
```
kay-jahnke: you are not checking the args or kwargs, how can you than determine that no keyword was given?

Neverhorst:
@kay-jahnke: yeah, the test logic got accidentally dumped too early since I want to get rid of all the testing stuff before merging.
If you want it you can check it out from an earlier version of this branch but it contained nothing of interest.

Neverhorst:
Ok, please check again. As we discussed I introduced more "draconic" constraints on the interface parameter in the `Connector` object.
Exceptions are raised when you try to connect to a module whose class name or base classes do not match `Connector.interface`.
I fixed all initialization calls for `Connector` in all modules within _default.cfg_ but this will potentially break any other module (especially modules connecting to non-interface classes)

PR #544, opened 2019-08-19
bug fix to forced sequence mode

kay-jahnke:
Small fix to only print a debug when the default sequence was successfully created. This bug was causing an additional error message, when the predefined function returned only empty lists.

## Motivation and Context

One can write predefined functions, that only return empty lists (so no waveforms at all). For these cases the load failed, because it assumed that the default sequence was successfully created. Now an if-clause catches this error.

As this is a bug fix to a new feature, I would leave it out of the changelog.

## How Has This Been Tested?

Linux 5.1.18-1-MANJARO on dummy config with different predefined functions not returning anything, some not even having a "name".

## Types of changes

## Checklist:

### DIFF
--- a/hardware/pulser_dummy.py
+++ b/hardware/pulser_dummy.py
@@ -45,6 +45,7 @@ class PulserDummy(Base, PulserInterface):
     _modtype = 'hardware'
 
     activation_config = StatusVar(default=None)
+    force_sequence_option = ConfigOption('force_sequence_option', default=False)
 
     def __init__(self, config, **kwargs):
         super().__init__(config=config, **kwargs)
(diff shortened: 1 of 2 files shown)


hardware/pulser_dummy.py
```diff
@@ -221,7 +221,7 @@ def get_constraints(self):
         activation_config['config9'] = frozenset({'a_ch2', 'a_ch3'})
         constraints.activation_config = activation_config
 
-        constraints.sequence_option = SequenceOption.OPTIONAL
+        constraints.sequence_option = SequenceOption.FORCED
```
Neverhorst: Maybe not a good idea. How about making it a `ConfigOption` for dummy.

kay-jahnke: good point, I like it.

Please check again.

PR #546, opened 2019-08-27
Scanner aberration interfuse

alrik-durand:
## Description
I've added an interfuse to transform the XY coordinate system of the confocal scanner to another by using two variable polynomes.

## Motivation and Context
Working off axis, for example with a steering mirror, can induce aberrations. There is then no linear mapping between the voltage applied on the hardware and distance on the sample. With this interfuse a bit of work, one can compute the transformation that counter these efects.

## How Has This Been Tested?
I've used in on our setup with the NI card 6323. It work nicely.

## Types of changes

## Checklist:


alrik-durand:
Thanks you for your review,
I've removed the config file and renamed the module, sorry for not being creative, I like the name you suggested !

PR #549, opened 2019-09-17
fixed: ramsey from list now accepts tau lists given as strings 

timoML:
## Description
Fix generation error in Ramsey from list in the predefined methods.

## Motivation and Context
So far generation only worked when providing a real Python list, eg. from a jupyter notebook. Now it works also from gui. 

## How Has This Been Tested?
Sequence generation from gui. Works on setup3 & dummy.

## Types of changes

## Checklist:

### DIFF
--- a/core/util/helpers.py
+++ b/core/util/helpers.py
@@ -275,3 +275,32 @@ def in_range(value, lower_limit, upper_limit):
     if value < l_limit:
         return lower_limit
     return value
+
+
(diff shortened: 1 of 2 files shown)


timoML:
Ok, I agree that string parsing should be avoided if possible and think float() casting is applicable here.
I would stick with your suggestion except for:
- square brackets should not lead to a raise
- non string inputs to generate_ramsey_from_list() should be possible to keep compability with existing notebooks

PR #553, opened 2019-11-26
Odmr exclusive analog

kay-jahnke:
So far whenever ODMR was performed with the NI-Card, a digital counter channel had to be selected, even if it was not needed. This has been upgraded and the NI-Card now supports ODMR with analog channels only.

## Description

The NI-Card has been modified to check if an analog channel was configured and only run digital tasks when really needed. You can achieve exclusive analog ODMR by either leaving the list of photon_sources or scanner_counter_channels empty or by completely omitting these two options in your config.

## How Has This Been Tested?

Tested on Hardware with Win10 64x.

## Types of changes

## Checklist:


hardware/national_instruments_x_series.py
```diff
@@ -141,6 +140,10 @@ def on_activate(self):
         self._oversampling = 0
         self._lock_in_active = False
 
+        self._photon_sources = self._photon_sources if self._photon_sources is not None else list()
+        self._scanner_counter_channels = self._scanner_counter_channels if self._scanner_counter_channels is not None else list()
+        self._scanner_ai_channels = self._scanner_ai_channels if self._scanner_ai_channels is not None else list()
+
```
Neverhorst: default value is an empty list, so why check for `None`? The user would have to explicitly declare `None` in the config for this to trigger.

hardware/national_instruments_x_series.py
```diff
@@ -99,13 +99,12 @@ class NationalInstrumentsXSeries(Base, SlowCounterInterface, ConfocalScannerInte
     _counter_ai_channels = ConfigOption('counter_ai_channels', [], missing='info')
```
Neverhorst: Maybe replace `[]` with `list()` as well here... uniformity and such...

hardware/national_instruments_x_series.py
```diff
@@ -165,6 +168,14 @@ def on_activate(self):
...
         """ Shut down the NI card.
         """
+        self._stop_analog_output()
+        # clear the task
+        try:
+            daq.DAQmxClearTask(self._scanner_ao_task)
+            self._scanner_ao_task = None
+        except:
+            self.log.exception('Could not clear AO Out Task.')
+
```
Neverhorst: Isn't that covered with `self.reset_hardware()`? If not, it should be.

hardware/national_instruments_x_series.py
```diff
@@ -1495,49 +1506,67 @@ def set_up_odmr(self, counter_channel=None, photon_source=None,
         else:
             my_clock_channel = self._scanner_clock_channel
 
-        if counter_channel is not None:
-            my_counter_channel = counter_channel
-        else:
-            my_counter_channel = self._scanner_counter_channels[0]
+        if len(self._scanner_counter_channels) > 0 and len(self._photon_sources) > 0:
```
Neverhorst: if you want to check against empty lists (or tuples, dicts etc.) you can also just ask `if first_list and second_list:`.

hardware/national_instruments_x_series.py
drogenlied: I can answer this one: `None` can occur when not adding any list elements after the key word in the config file.
As soon as there is a new line with a `-`, a list is returned.

Neverhorst: It's just extra code, but ok... in the unlikely case someone declares a config option without giving a value this specific module will cope with it... ;)

By the way: This piece of code would also be rendered obsolete if checks for empty channel lists would be done by converting it to `bool` instead of checking `len == 0`. (see below)

hardware/national_instruments_x_series.py
kay-jahnke: We thought so too, but we had the case where if we are also accessing the same hardware with another module and then call `reset`, sometimes the old tasks linger around.
So I though better save then sorry.

hardware/national_instruments_x_series.py
```diff
@@ -1919,7 +1911,9 @@ def close_odmr(self):
         return retval
 
     def get_odmr_channels(self):
-        ch = [self._scanner_counter_channels[0]]
+        ch = list()
+        if self._scanner_counter_channels:
+            ch = ch.append(self._scanner_counter_channels[0])
```
Neverhorst: This will raise an error during activation.

This line will leave `ch == None` due to `list.append()` returning `None`.
Leave out the new assignment to `ch` and just append.

PR #555, opened 2019-12-04
Slow counter overhaul

Neverhorst:
## Description
Added a whole new tool chain (GUI, logic, interface, hardware) for "slow counting".

In the process I also deleted some testing/template modules that somehow made their way into master... but should not have...

## Motivation and Context
The current time series counter is very limited regarding data throughput and mixed signal capabilities.
For high data rates it will crash the NI card and dependent modules.
Also the oversampling and averaging is very inconsistent.
I implemented a wholesome replacement for "slow counting" by using the new modules `TimeSeriesGui`, `TimeSeriesReaderLogic` and `NIXSeriesInStreamer`.
The logic and hardware interface are more versatile and should be suited for any use case where continuous data streaming from a buffered device with variable read size is needed.
Also the counter/time trace GUI offers settings to switch channels physically and/or virtually on and off.

The old counter modules are still present but loading the respective GUI will cause a deprecation warning to pop up in order to encourage users to switch to the new modules.
I also intend to include the pop-up notification as simple tool in core... but in a different PR. 

## How Has This Been Tested?
On Win8.1 x64 with NI USB-6343 hardware and also with new dummy module.

## Types of changes

## Checklist:


Neverhorst:
Thank you @kay-jahnke.
I tried to reproduce this error on dummy config with 1kHz data rate, 4 times oversampling (=4kHz sampling rate) and 20s trace window.
Started sampling a large hahnecho waveform and pressed "load all modules". Everything stutters and it takes ages but in the end... no error.
Can you give a more detailed way to reproduce this error?

alrik-durand:
Hi, 

Sorry for the late message ! Could I suggest some modification to the interface architecture ?
As I mentioned in the issue #586, it would be great to have some kind of guidelines for the interface design, so all the interfaces (at least the new ones) follow a similar style.
I personally have some preferences, but i would be happy with any unified style.

I can make some suggestion/change to this code if you are interested but don't have time for this !

What are your thoughts on this ?

Alrik

PR #556, opened 2019-12-10
Refocus while ODMR and pulsed

alrik-durand:
## Description
I've added an optional connector to POI logic so that both ODMR and pulsed measurement can be paused before starting the refocus sequence

## Motivation and Context
Long acquisition like ODMR and pulsed acquisition suffer from drifts. Right now there is no way to do what this small PR does.

## How Has This Been Tested?
ODMR was tested on a real setup. Pulsed only on a dummy.

## Types of changes

## Checklist:


alrik-durand:
Thank you for your review,
I agree that the link to pulsedmeasurement is unnecessary and too measurement specific, I will remove it entirely. Regarding ODMR, the is less clear to me. I don't know all the kind of measurement that are done with this module, but I don't see why (at least for the one I know) you would not want to pause the ODMR for the refocus, as it will for affect the result for sure ? As it is an optional connector, one do not need to drag ODMR logic into it if not necessary.
Can you help me understand better ?

drogenlied:
There is infrastructure to deal with the problem of starting/stopping measurements and executing actions, but nobody seems to use that and everyone I know of just writes notebooks.

Due to this disuse, the example task configuration was broken, so if you want to try this out, look at the `fix-refocus-task` branch.

The modules are `logic/taskrunner.py` and `gui/taskrunner/taskgui.py` the associated tasks are stored in `logic/tasks`.

Rather than adding lots and lots of optional connectors for each measurement that could possibly happen, adding optional connector to `tasklogic` in  the POI manager and the configuration of some tasks could solve this in a much cleaner way.

Further steps away from notebook scripting would mean to finish `logic/automation.py`, which had the goal to make executable sequences and loops out of tasks for repetitive measurements.
Sadly, this never got done due to lack of interest as notebook scripting was seen as good enough.

alrik-durand:
Thank for you answers,
I would love to look into automation.py but it's a bit hard from what is in the logic.
Is the file auto.cfg available somewhere ? Maybe some example would help.
We're also generally using notebooks here but they are way harder to share and maintain.

PR #560, opened 2019-12-27
Pulsed timetrace analysis

alrik-durand:
## Description
I've added a new tab to pulsed measurement, with features to plot and analyse the timetrace directly.
I've also added checkboxes to save selectively pulses extracted data or timetrace data, to give access to this feature pre-existing in logic.
I have also created a utils/curve_methods with general function I needed and believe to be more general than this module. I was hoping to start a kind of library for function used generally in data analysis of Qudi's files, but this may need it's own thread.

## Motivation and Context
Pulsed measurement is very useful to extract spin state information out of pulses, but lacked features to just analyse the timetrace, giving helpful information on the rise or fall times of pulses, extracting lifetimes or other features of time resolved experiments.

## How Has This Been Tested?
I've tested this with dummy modules, I've been as careful as possible to not break any pre-existing behavior of the pulsed module. I've checked that it work as I intend on my personal linux machine.

## Types of changes

I've ticked breaking change as I'm worried the *do_fit* and related functions have a slightly different prototype, with a string instead of a boolean. Other than that it should work fine.

## Checklist:

### DIFF
--- a/core/util/curve_methods.py
+++ b/core/util/curve_methods.py
@@ -1,0 +1,102 @@
+# -*- coding: utf-8 -*-
+"""
+This file contains Qudi methods for data curve manipulation, either just Y, (X, Y) or (X, [Y1, Y2, ... YN]
+
+---
+
+Qudi is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+Qudi is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with Qudi. If not, see <http://www.gnu.org/licenses/>.
+
+Copyright (c) the Qudi Developers. See the COPYRIGHT.txt file at the
+top-level directory of this distribution and at <https://github.com/Ulm-IQO/qudi/>
+"""
+
+import numpy as np
+
+
+def rebin(y, ratio, do_average=False):
+    """ Rebin a 1D array the good old way.
+
+    @param (1d array) y : The Y vector to rebin
+    @param (int) ratio: The number of old bin per new bin
+    @param (bool) do_average: Optional parameter to take the average instead of the sum in the process
+
+    @return (1d numpy array) : The array rebinned
+
+    This rebbining method does not use any dangerous smoothing method. It just simply adds bin.
+
+    The last values will be dropped if the sizes do not match
+    """
+    y = np.array(y)
+    ratio = int(ratio)
+    length = (len(y) // ratio) * ratio
+    y = y[0:length]
+    y = y.reshape(length//ratio, ratio)
+    if do_average:
+        return y.mean(axis=1)
+    else:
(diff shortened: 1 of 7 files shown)


Neverhorst:
As much as I can understand the temptation to turn the pulsed measurement "thingy" into a general purpose data playground, I must opt against this PR.
Current pulsed measurement is already way too bloated and in-transparent.

The idea of qudi as a project is to provide a highly modular(!) framework and library for scientists to use and tailor to their needs even without advanced programming knowledge. So one important key feature is simplicity and readability of code for experimentation modules.
Pulsed measurement as a toolbox is straining this principle quite heavily already and should be divided into more logic modules.
So you should only improve/expand existing experimentation (logic) modules if it helps them perform their intended task better.
For a new class of experiments (like analyzing time trace data in a totally different way for e.g. single shot readout) it is mandatory to write a new logic module.
If you can re-use some features from pulsed measurement (like `SequenceGeneratorLogic`) it is perfectly fine to connect to these modules.

The overhead of writing new modules is quite small.

If people want to discuss this issue further, please feel free to do so here (and re-open PR if necessary).

alrik-durand:
I would indeed like to discuss this.
But first thank you for your feedbacks.

I've been frustrated many time by the impossibility to save an image of the raw timetrace from pulsed measurement. That is why I think it helps perform the intended task better. 
I understand your point about the complexity of pulsed measurement, but in fact I would say maybe the time trace analysis should have been part of the "basic package" and complex sequences analysis/pulse extraction should be constructed on top of it, in another module ?

I have to admin I would like this to be merged, but I'm open for constructive discussions :)

Neverhorst:
Ok let's see:

- saving image files of data graphs was only ever intended as "thumbnail" for the measurement outcome. If you want to use qudi as plotting front-end for publication/presentation, please write your own plotting module. You can also have a look at `gui.simpledatagui`.
There are at least as many opinions about how data plots should look like as there are researchers using qudi.
In fact most complaints I receive about saving data in qudi originates from spamming image files all over the place by default which also reduces the data save performance by a factor of up to 100.
That is why there are tendencies to reduce the places where images are created.
If you just want some snapshot image of (a zoom-in of) the raw time trace you can use the plot in the "pulse extraction" tab and export an image with the right-click context menu (as with almost any plot).

- One of the reasons why pulsed measurement is such a complex module is the capability to dynamically add raw data processing extensions. They are called extraction methods and analysis methods. 
For extraction methods you can simply drop in a "black box function" that takes any number of user inputs (automatically generated in GUI) as well as the raw time trace and returns a 2D numpy array containing basically anything.
For analysis methods you can do something similar by taking previously mentioned 2D array and returning data points shown in the analysis tab.
Refer to `how_to_add_extraction_methods` and `how_to_add_analysis_methods` in documentation.
This feature is very flexible so if you can not use it for your pulsed measurement data analysis purposes, this is most likely the wrong module for your experiment.

- If the above feature can not help you, then you literally just need the `SequenceGeneratorLogic` and a raw time trace from a counting device (both available in qudi as separate modules) and write some custom logic to process that data in the way you like.

alrik-durand:
Thank you again for all your advises. It may seems simple to you but it's really helpful to me !

I like you idea of using extraction and analysis methods to do the same thing, I have never looked in detail at this part of Qudi so I didn't even think to look into that. I will for sure. 

I'll try to do it this way and maybe submit it in another PR (it still may be useful to someone else)

Case closed !

PR #561, opened 2019-12-27
Pulsed trigger ratio

alrik-durand:
## Description
I have added an attribute of pulsed logic that indicate that proportion of the time the fast counting device was in *triggered and counting* state over the timescale of the acquisition.

## Motivation and Context
When we  generate a sequence, we often have to make sure the acquisition time is shorter than the sequence to not miss some trigger pulses. 
One strategy is to introduce a margin by which all acquisition are reduced, that is what that fastcomtec.py trigger_safety config option does for example. Sadly this tend to be forgotten over time by the different users, so instead I've tried adding an indicator of how efficient the acquisition is.

## How Has This Been Tested?
I have tested this with dummies on my linux PC.

## Types of changes

## Checklist:


alrik-durand:
Again, thank you for your feedback. 
I agree the FastComTec can sometimes be a pain. But I'm surprised you consider this a hardware specific problem : to me any hardware will have a problem of missing trigger if the measured length is exactly the programmed sequence length. That is why I though it would be relevant to every pulsed hardware. Maybe I'm missing something ?
I've only used FastComTec so far so my own experience if very hardware specific...

Neverhorst:
It really depends on the inner working principle of the device.
Some devices, like FastComTec, are "blind to triggers" until the specified counting time has elapsed; others will simply restart/jump to next count trace even though the old one has not been finished (last few bins will have zero events).

alrik-durand:
I did not respond to your main proposition in your message from April. The ratio between *sweeps x duration / elapsed_time* will depend on the application, it should not always be one, for example if you only want to record part of the sequence (which is often the case for long sequences) therefore a hardware module shouldn't raise an error nor even a warning just because this ratio is bellow a given number. This is something that is very sequence dependent.
I don't see how else to treat this issue so that the user have all the information at his disposal ?

logic/pulsed/pulsed_measurement_logic.py
```diff
@@ -1153,6 +1159,7 @@ def _extract_laser_pulses(self):
         self.raw_data = fc_data
         self.__elapsed_sweeps = info_dict['elapsed_sweeps']
         self.__elapsed_time = info_dict['elapsed_time']
+        self.__trigger_ratio = (self.__elapsed_sweeps * self.__fast_counter_record_length) / self.__elapsed_time
```
Neverhorst, line 1162: All the values to calculate this trigger ratio are already known to the fast counting hardware. So at least regarding calculating this ratio there is no need to involve the measurement logic.

Neverhorst (commented):
I completely understand your troubles.
However, bottom line is: Bad/Buggy hardware does not warrant a specific workaround in measurement logic. You need to find some way to handle this "error" in the hardware module.

As a proposal, maybe use a more suited figure of merit than the trigger ratio you are currently calculating? Even if you are not recording the entire pulse sequence, you would still have a fixed ratio/frequency of sweeps if no triggers are missing. Would monitoring this frequency and alerting the user if the deviation is greater or equal one counting length solve the issue?

logic/pulsed/pulsed_measurement_logic.py
alrik-durand: This is very true, I did not know most other hardware did not have a problem of dead time when I wrote this, which is my a approached it in a hardware agnostic perspective. If they don't then it's true this is quite useless !

To answer your previous message, the ratio does not fluctuate in time, because it's not a problem of buggy hardware.

The error I'm trying to prevent is when the user forget the dead time of the fast counter and send the next trigger too soon in the programmed sequence. If he does, he will have a constant 50% ratio, and this might help it prevent such error. That is what this PR is about. 

Another solution is we drop the PR and too bad for the PhD students who forget about the dead time !

PR #566, opened 2020-01-09
Pulsed refocus task

alrik-durand:
## Description
I've added a task that pause pulsed measurement, run laser_on, start a poi refind then goes back to original acquisition.

## Motivation and Context
Some long pulsed acquisition like T1 measurement have laser on only a short percentage of the time, causing issue tracking because there is no fluorescence to track. 
Following the discussion of #556, this is a solution for a particular need.

## How Has This Been Tested?
This was tested on dummies.

## Types of changes

## Checklist:


Neverhorst:
So far I see no problem with this PR. Unfortunately this code needs to be revisited once qudi has been released as package (using PySide2) since this will bring drastic changes to qudi core.

Since I have never used tasks, maybe @drogenlied wants to have a quick glance in here?

timoML:
Short update: I started to [port this](https://github.com/Ulm-IQO/qudi-iqo-modules/tree/refocus_tasks/src/qudi/tasks) to the new core. As it is a quite usefull mechanism to do all sorts of things before starting optimise(), we will merge this as soon as the optimiser becomes available again in the new core. This may take a while, as qudi-iqo-modules is still going through a major rework.

Sorry for the delay and for not preserving your commit history in the new repo (contact me if we should try to restore).

Issue #323, opened 2018-02-16
Save Logic doesnt let the user/logic choose which kind of figure format should be used when saving (eg. pdf/png/jpeg)


AndreaFilipovski:
## *What* is affected by this issue?
Any saving process that uses big data amount in figures (eg. 20MB) is saving multiple data formats which is computational  and time intensive without it being necessary to the user.
It would be more helpful if only wanted data formats are saved.

## *When* does this occur?
Every time a save is used

## *How* do we replicate the issue?
Save any figure with data using the save logic method using "save_data()" with plotfig not None using big amounts of data.

## Expected behavior (i.e. solution)
It should be possible to pass a variable to the method which defines into which kind of data format the figure is to be saved.
For it to be flexible I suggest a list where many formats (eg. like it is now done pdf and png) can be passed.
The corresponding part of the code for different figures could then only be executed if the corresponding file format was passed.
For this it might make sense to implement more data formats like svg or jpeg.

## Other Comments
To make it really flexible by introducing more file formats might be a bigger issues while it might not add a lot of usability. Therefore it might be sensible to choose the easier option to pass plotfig as before and add new variable fileformatplot = None. 
If this is passed the type passed (can only be pdf or png) will be used, if nothing is passed both will be plotted.
The advantage here that implementation is fast and not very error prone. 
The disadvantage is that it is still quite inflexible.

PR #572, opened 2020-01-19
Save improvements

johnjarman:
## Description
I wanted to share some of our changes to saving in qudi, in case they are useful.

### Saving confocal data no longer freezes other GUIs
Change confocal to use a similar method to ODMR for saving, i.e. using a logic module signal to free up the GUI thread.

These changes prevent blocking of the GUI thread during saves by modifying the `save_xy_data` and `save_depth_data` methods in confocal_logic.py. They now have a new kwarg `block` (default `True`). 
If set to `False`, these methods will return immediately after emitting a Qt signal, which is queued in the confocal logic module's thread. The signals are connected to renamed methods `_save_depth_data` and `_save_xy_data`. In the case where `block` is not specified or set to `True`, the modified methods behave exactly as before to avoid breaking code that depends on this blocking for synchronisation.

The confocal GUI is updated to call `save_xy_data` and `save_depth_data` with `block=False`, so that the GUI thread is released.

If `block=False`, the confocal logic thread is blocked instead of the calling thread during the save. Any signals connected to slots in the logic module are enqueued and processed after the save completes, so repeated calls to `save_xy_data` (for example) will lead to multiple saves occurring sequentially. Likewise if there is a running scan, the scan effectively pauses while the save is running. Previously, the confocal logic thread was not blocked during saves and so data could update while the save was running (e.g. if a scan was in progress)

Leaving the GUI unblocked can lead to counter-intuitive behaviour if the confocal GUI is used during saving, since user commands are also queued up and executed in sequence after the save completes. Hence, a new window-modal dialog (`ui_save_dialog.ui`) is added and displayed during saving, to block the entire confocal GUI window. A new signal `signal_save_started` is used to trigger display of this dialog, and the old `data_saved` signals are used to dismiss it when it's done. The rest of the Qudi GUI is not blocked during this operation, allowing interaction with other modules. The dialog is provided in the `qtwidgets` directory for possible application to other modules with saves or other blocking ops that take more than a second or so.

### Fix daily folder bug in save_logic.py
save_logic.py (on our Windows systems at least) always says it's creating a new daily directory whenever data is saved. This commit cleans up the logic and seems to fix this bug.

### Add config options for PDF & PNG to save_logic
We prefer not to save a PDF since it clutters up the save directory and we never use it. Therefore I've added new config options `save_pdf` and `save_png` that enable/disable each output file (default to enable).

## Motivation and Context

* Config options for PDF & PNG partly resolves Issue #323.
* I think implementing a save that blocks the calling logic module but not the entire GUI gets around some concerns in [PR 393](https://github.com/Ulm-IQO/qudi/pull/393#issue-208244406). This is not a general solution, but it is also extremely limited in scope and solves a concrete problem which is that the entire GUI, not just the confocal module, is frozen during long-running save operations of confocal data.
* The changes are similar to how the ODMR module already works.
## How Has This Been Tested?
Tested on Python 3.6.5 and 3.7.6 on Windows 10, with dummy config and on lab setup
* Saving confocal data (depth and xy scan) both while scanning and while not scanning; confirm that other GUI windows are not blocked but confocal GUI is blocked by modal dialog
* Saving with every combinaton of new config options (times: both: 1.3 s, png: 0.98 s, pdf: 0.56 s, neither: 0.01 s)
* Deleting save directory and subdirectories to confirm (re-)creation works
* Using iPython console to run `scannerlogic.save_data_xy()` &  `scannerlogic.save_data_depth()` to confirm old behaviour (i.e. blocking) is retained when new kwarg `block` is not specified
* Using iPython console to run `scannerlogic.save_data_xy(block=False)` repeatedly to check for weird behaviour
[Screenshot of save in progress](https://imgur.com/a/mpNIoCk)
## Types of changes

## Checklist:


johnjarman:
Thank you for taking the time to review this. I've implemented most changes as requested.
However, I think it's potentially useful for users to be able to dismiss the save dialog in case there is an exception raised during the save process that prevents `signal_xy_data_saved` being emitted. The dialog is useful as visual feedback and protection against accidental errors in its current state, and I think that's probably good enough. I could add try..finally blocks to save_logic and remove the close button if preferred, though.

johnjarman:
Rebased on to master - should be ready to merge 👍

PR #573, opened 2020-01-20
Multiple ranges odmr

Geegee:
Allow adding new ranges in the ODMR GUI. For this one can add and remove
Spin Boxes to enter the different measurement ranges. 

## Description

The ODMR gui and logic has been adjusted. One can fit each range individually.
Select the matrix corresponding to the range. 
simplescreenrecorder-2020-01-20_15 35 56

## Motivation and Context

It can save measurement time. 
## How Has This Been Tested?

Tested with dummy hardware and on widefield setup. 

Running ODMR with AWG70k and camera.

The changes affect ODMR logic and gui.
## Types of changes

## Checklist:

- [x ] I have checked that the change does not contain obvious errors (syntax, indentation, mutable default values).


liuyanulm:
This is a very nice new feature. I have done the test on dummy. It works well.

NicolasStaudenmaier:
Hi Gerhard, this looks like an interesting new feature that will come in handy. Can you please update to the current master version? Then I will properly test and check it.

NicolasStaudenmaier:
And a first comment. The save_odmr_data function in the odmr_logic needs to be revised. There are still the old parameters mw_start/stop/step. When saving the odmr data it causes an error.

Geegee:
Thanks a lot Nico for the suggestions. I totally forgot about the save function and updated the branch to master. 
I hope the result is satisfactory now.

gui/odmr/odmrgui.py
```diff
@@ -366,6 +476,179 @@ def _menu_settings(self):
...
+            gridLayout.removeWidget(object_dict[object_name])
+            del self._mw.odmr_control_DockWidget.__dict__[object_name]
+
+        starts = self.get_frequencies_from_spinboxes('start')
+        stops = self.get_frequencies_from_spinboxes('stop')
+        steps = self.get_frequencies_from_spinboxes('step')
+        power = self._mw.sweep_power_DoubleSpinBox.value()
+        self.sigMwSweepParamsChanged.emit(starts, stops, steps, power)
+
+        self._odmr_logic.range_to_fit -= 1
```
NicolasStaudenmaier, line 584: If range_to_fit is 0 and ranges are reduced this might set the value to a negative one causing an error eventually. Simply reset to 0 each time or check whether it is already 0.

NicolasStaudenmaier, line 584: If range_to_fit is zero this might set a negative value and will eventually cause an error when fitting. Simply reset to 0 each time or check whether is it already 0.

logic/odmr_logic.py
```diff
@@ -201,12 +207,35 @@ def sv_get_fits(self, val):
...
+        final_freq_list = []
+        for mw_start, mw_stop, mw_step in zip(self.mw_starts, self.mw_stops, self.mw_steps):
+            freqs = np.arange(mw_start, mw_stop + mw_step, mw_step)
+            final_freq_list.extend(freqs)
+            self.frequency_lists.append(freqs)
+
+        if self.final_freq_list:
+            self.odmr_plot_x = np.array(self.final_freq_list)
+        else:
+            self.odmr_plot_x = np.array(final_freq_list)
```
NicolasStaudenmaier, line 222: For the the if phrase of a list is causing an error. What do you want to check here, whether the list is empty? In that case you can use "if len(self.final_freq_list)".

logic/odmr_logic.py
```diff
@@ -387,38 +416,51 @@ def set_cw_parameters(self, frequency, power):
...
-    def set_sweep_parameters(self, start, stop, step, power):
+    def set_sweep_parameters(self, starts, stops, steps, power):
         """ Set the desired frequency parameters for list and sweep mode
 
-        @param float start: start frequency to set in Hz
-        @param float stop: stop frequency to set in Hz
-        @param float step: step frequency to set in Hz
+        @param float starts: list of start frequencies to set in Hz
+        @param float stops: list of stop frequencies to set in Hz
+        @param float steps: list of step frequencies to set in Hz
```
NicolasStaudenmaier, line 424: @param list starts/stops/steps

logic/odmr_logic.py
```diff
@@ -747,6 +817,7 @@ def _scan_odmr_line(self):
 
             # Set plot slice of matrix
             self.odmr_plot_xy = self.odmr_raw_data[:self.number_of_lines, :, :]
+            key = 'Matrix range: {}'.format(self.matrix_range)
```
NicolasStaudenmaier, line 820: This variable isn't used anywhere.

logic/odmr_logic.py
```diff
@@ -454,41 +496,69 @@ def mw_sweep_on(self):
...
+                                     'Lowering resolution to fit the maximum length.')
+                    mw_step = np.abs(mw_stop - mw_start) / (limits.list_maxentries - 1)
+                    self.sigParameterUpdated.emit({'mw_steps': [mw_step]})
+
+                sweep_return = self._mw_device.set_sweep(
+                    mw_start, mw_stop, mw_step, self.sweep_mw_power)
+                mw_start, mw_stop, mw_step, self.sweep_mw_power, mode = sweep_return
+
+                param_dict = {'mw_starts': [mw_start], 'mw_stops': [mw_stop],
+                              'mw_steps': [mw_step], 'sweep_mw_power': self.sweep_mw_power}
```
NicolasStaudenmaier, line 559: 1) The data for the lists self.frequency_lists and self.freq_final_list is not set here. I think this causes the error when working in SWEEP mode.

logic/odmr_logic.py
```diff
@@ -454,41 +496,69 @@ def mw_sweep_on(self):
...
 
-            param_dict = {'mw_start': self.mw_start, 'mw_stop': self.mw_stop,
-                          'mw_step': self.mw_step, 'sweep_mw_power': self.sweep_mw_power}
+                num_steps = int(np.rint((mw_stop - mw_start) / mw_step))
+                end_freq = mw_start + num_steps * mw_step
+                freq_list = np.linspace(mw_start, end_freq, num_steps + 1)
+                if np.abs(mw_stop - mw_start) / mw_step >= limits.list_maxentries:
+                    self.log.warning('Number of frequency steps too large for microwave device. '
+                                     'Lowering resolution to fit the maximum length.')
+                    mw_step = np.abs(mw_stop - mw_start) / (limits.list_maxentries - 1)
```
NicolasStaudenmaier, line 514: This is using the length of each range individually but the total length of all ranges should be checked here. So check the total length somewhere and then either reduce the resolution or raise an error.

NicolasStaudenmaier (changes requested):
Hi Gerhard, I have reviewed your pull request and some minor changes are necessary. Altogether it looks good and is mostly working well. However, some things need to be reviewed. Here is a summary and you can also find my comments in the code.

1) If you are running with SWEEP mode generation of the matrix diagram is causing an error. I suppose this is because when turning on the MW (mw_sweep_on), the data for 'frequency_lists' and 'final_freq_list' is not set. 
2) In _initialize_odmr_plots() there is an if-phrase of a list. For me that is causing an error. If you just want to check that the list is not empty use "if len(list):". 
3) In the GUI module the chosen range on which to perform a fit us reduced by 1 each time you remove a range (line 584 in remove_ranges_gui_elements_clicked()). This might set the chosen range to a negative value and eventually causes an error. 
4) When starting an ODMR scan the length of each range is checked to be smaller than the max length given by the hardware constraints. One should check that the length of all ranges together doesn't exceed the maximum. So you have to check that at a different point and then either reduce the number of points or raise an error. 

Some minor things:
5) Docstring of "set_sweep_parameters": the parameters start, step, stop in are still labeled as floats. 
6) You can do some code cleaning in the logic file. Especially removing excess blank lines. 
7) As this pull request adds some nice functionality and may cause breaks for someone who is accessing the ODMR module by external scripts, please add the new functionality to the changelog file.

gui/odmr/odmrgui.py
Geegee: This can not happen as at the top of the function it is checked that the number of ranges can not fall below 1.

Geegee:
Thanks again for taking the time to look over the code and making good points! 
I hope I could address all points.

gui/odmr/odmrgui.py
NicolasStaudenmaier: Yes, this assures that one range always remains. But the problem I mean is that the parameter `range_to_fit` can fall below zero. Assume that you have three ranges for example and you have `range_to_fit = 0`. If you remove a range now this function call will set `range_to_fit = -1` and if you press the fit button you get an error.

logic/odmr_logic.py
```diff
@@ -208,15 +208,15 @@ def sv_get_fits(self, val):
...
         final_freq_list = []
         for mw_start, mw_stop, mw_step in zip(self.mw_starts, self.mw_stops, self.mw_steps):
             freqs = np.arange(mw_start, mw_stop + mw_step, mw_step)
             final_freq_list.extend(freqs)
             self.frequency_lists.append(freqs)
 
-        if self.final_freq_list:
+        if type(self.final_freq_list) == list:
+            self.final_freq_list = np.array(final_freq_list)
+        if self.final_freq_list.all():
```
NicolasStaudenmaier, line 219: Can you please specify what the purpose of this if statement is? I think it will give `True` in almost all use cases as `False` is obtained only if there is a zero in the np.array. It also yields `True` when the array is empty.

gui/odmr/odmrgui.py
```diff
@@ -366,6 +478,179 @@ def _menu_settings(self):
...
+        steps = self.get_frequencies_from_spinboxes('step')
+        power = self._mw.sweep_power_DoubleSpinBox.value()
+        self.sigMwSweepParamsChanged.emit(starts, stops, steps, power)
+
+        # in case the removed range is the one selected for fitting right now adjust the value
+        self._odmr_logic.ranges -= 1
+        max_val = self._odmr_logic.ranges
+        if self._odmr_logic.range_to_fit == max_val:
+            self._odmr_logic.range_to_fit = max_val
+            self._mw.fit_range_SpinBox.setMaximum(max_val)
```
NicolasStaudenmaier, line 591: `max_val = self._odmr_logic.ranges - 1`
`if self._odmr_logic.range_to_fit > max_val:`
line 591 should not be indented

logic/odmr_logic.py
```diff
@@ -454,41 +489,64 @@ def mw_sweep_on(self):
...
+                final_freq_list.extend(freq_list)
+                # self.frequency_lists.append(freq_list)
+
+                used_starts.append(mw_start)
+                used_steps.append(mw_step)
+                used_stops.append(end_freq)
+
+            final_freq_list = np.array(final_freq_list)
+            if len(final_freq_list) >= limits.list_maxentries:
+                self.log.error('Number of frequency steps too large for microwave device.')
```
NicolasStaudenmaier, line 515: After the error message you have to properly leave the function. The following code (same as at the end of the function) will do the job:
`mode, is_running = self._mw_device.get_status()`
`self.sigOutputStateUpdated.emit(mode, is_running)`
`return mode, is_running`

NicolasStaudenmaier (commented):
I have left another two short comments. When these are fixed everything is ok. Sorry for being so finical ;)

gui/odmr/odmrgui.py
```diff
@@ -585,10 +585,11 @@ def remove_ranges_gui_elements_clicked(self):
...
         # in case the removed range is the one selected for fitting right now adjust the value
         self._odmr_logic.ranges -= 1
-        max_val = self._odmr_logic.ranges
-        if self._odmr_logic.range_to_fit == max_val:
+        max_val = self._odmr_logic.ranges - 1
+        self._mw.fit_range_SpinBox.setMaximum(max_val)
+        if self._odmr_logic.range_to_fit > max_val:
             self._odmr_logic.range_to_fit = max_val
-            self._mw.fit_range_SpinBox.setMaximum(max_val)
+
```
NicolasStaudenmaier: You have to set the new maximum for the SpinBox after this if-phrase. Otherwise this is causing some strange behavior for the odmrlogic's `range_to_fit`.

NicolasStaudenmaier (approved):
Looks everything good now. Thanks for your work, Gerhard!

PR #577, opened 2020-02-20
Fastcomtec 7889 & DMA mode

alrik-durand:
## Description
This PR adds compatibility with fastcomtec model 7889 through a config option in 7887 hardware file. 
This PR also adds the option to set DMA mode to true by default with another config option.
Finally, get_current_sweeps now uses the start number and not the sweep numbers.

## Motivation and Context
1. It's nice when you have a 7889 card.
2. 7889 model have been seen to miss some triggers when not in DMA mode.
3. Also when it does, the number of start and sweeps become different, for a reason still unclear to me. Maybe I'm wrong but from my test, it seem the relevant number is the start number.
The only reason I'm using the sweep number is to normalize the timetrace to get counts per second that matches the slow counter. When *starts* and *sweeps* differ, the start number is the one that produce an count per second that matches slow counter.

## How Has This Been Tested?
I've tested this on 7889. I have not tested this on a 7887 but there should not be any change for it.

## Types of changes

## Checklist:


kay-jahnke:
This PR looks fine to me but someone should please test it on a Fastcomtec 7887 hardware.
@Neverhorst @SimonSchmitt88 @timoML @MathiasMetsch @Yawghmoth Do any of you have the 7887 and could test please.

Neverhorst:
I don't have one and never had one either.

timoML:
Tested on setup3 with p7887. Runs Rabi as expected.

Neverhorst (approved):
Since this PR has been open for quite some time, I think we won't get more tests from people with FastComTec hardware. Thank you, @timoML for testing. I think we should merge.

PR #581, opened 2020-03-20
Model/view for GUI

alrik-durand:
## Description
I've added a feature to easily link the state of a logic model to the state of a GUI view. I've also updated counter GUI to demonstrate how to use it.

## Motivation and Context
Adding just one spinbox in GUI and connecting it to logic can be a pain. Activation, deactivation, updating logic from GUI and GUI from logic all have to been done by hand at different part of the code. One usually use copy/paste a lot, never a good thing in a program !
Here the idea is to factor code by rendering explicit the concept of model and view.

Also the code of counter GUI needed some makeup !

## How Has This Been Tested?

I've tried testing as much as possible with counter GUI because adding abstraction can mess up things. I hope I have eliminated the last bug. This is tested on windows with dummies, no rpyc.

## Types of changes

## Checklist:

### DIFF
--- a/core/gui.py
+++ b/core/gui.py
@@ -20,6 +20,7 @@ top-level directory of this distribution and at <https://github.com/Ulm-IQO/qudi
 """
 
 import platform
+import types
 from qtpy.QtCore import QObject
 from qtpy.QtWidgets import QApplication
 from qtpy.QtGui import QIcon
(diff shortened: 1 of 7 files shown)


Neverhorst:
Have you had a look at https://github.com/Ulm-IQO/qudi/blob/master/core/mapper.py ?

alrik-durand:
No I did not know about that, but it looks very nice. Qudi will never cease to surprise me !

I guess I can close this PR, is it normal that no GUI module use this though ?

Neverhorst:
From my personal chats with qudi programmers/users it seems like most people prefer to keep the full flexibility that Qt offers in signalling and event handling rather than committing to a fixed behaviour by using the mapper.

alrik-durand:
I see why it is easier in some cases, I think some people (like me) were simply unaware of this feature. Are you ok if I try to implement it on at least one module as a showcase ? By doing this and adding something in the docs, maybe this nice feature would be used a bit more ?

Neverhorst:
> I see why it is easier in some cases, I think some people (like me) were simply unaware of this feature. Are you ok if I try to implement it on at least one module as a showcase ? By doing this and adding something in the docs, maybe this nice feature would be used a bit more ?

Sure

alrik-durand:
Ok, I'll close this one and open a new one for what we discussed.

PR #585, opened 2020-04-07
fixed SSL authentication of module server

tobiasgehring:
The SSL authentication of the module server was completely broken and potentially never worked. This pull request fixes it.

## Description
Now SSL can be used for authentication.

## Motivation and Context
Unencrypted and not authenticated connections shouldn't be used. But Qudi's implementation of SSL authentication was broken.

## How Has This Been Tested?
Qudi module server <-> manual rpyc connection on two Windows 10 machines.

## Types of changes

## Checklist:

### DIFF
--- a/core/manager.py
+++ b/core/manager.py
@@ -155,8 +155,18 @@ class Manager(QtCore.QObject):
                                 'port', 12345)
                             certfile = self.tree['global']['module_server'].get(
                                 'certfile', None)
+                            if (certfile is not None) and not os.path.isabs(certfile):
+                                certfile = os.path.abspath(os.path.join(self.configDir, certfile))
                             keyfile = self.tree['global']['module_server'].get('keyfile', None)
-                            self.rm.createServer(server_address, server_port, certfile, keyfile)
+                            if (keyfile is not None) and not os.path.isabs(keyfile):
+                                keyfile = os.path.abspath(os.path.join(self.configDir, keyfile))
+                            cacertfile = self.tree['global']['module_server'].get('cacertfile',
+                                                                                  None)
+                            if (cacertfile is not None) and not os.path.isabs(cacertfile):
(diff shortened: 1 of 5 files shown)


Neverhorst:
Is the version change for `rpyc` neccessary? And if yes, does it only apply to windows 10? Just asking because if it is neccessary for this PR to work, it would be a breaking change that requires a fresh install after updating.
And does the above also apply for the conda environments for other OS?

tobiasgehring:
Turns out it isn't. I just tested it with a fresh anaconda environment on Windows 10 and rpyc 4.0.2. Turns also out that 4.1 is not backwards compatible to 4.0, so server and client running on  two different versions cannot talk to each other. So I reverted the version bump.

Neverhorst:
> Turns out it isn't. I just tested it with a fresh anaconda environment on Windows 10 and rpyc 4.0.2. Turns also out that 4.1 is not backwards compatible to 4.0, so server and client running on two different versions cannot talk to each other. So I reverted the version bump.

Ok, thank you for checking.
And yes, you were right about encrypted connections never really working with remote modules. I also agree that simple two-way authentication is the most user-friendly way to go.
Looks good so far. I will test it on different machines as soon as I find the time.

tobiasgehring:
Any news and this? Can it be merged?

Neverhorst:
Somehow I could not get this to run properly on windows (8.1 and 10 with openSSL for windows).
A colleague of mine also encountered the same problem. I strictly followed your documentation with using a self-signed certificate. Maybe I'm missing something obvious?

Error on client:
client_error

Error on server:
server_error

Neverhorst (approved):
Well, since certified connections were not working before and now it seems to work (for some at least), I would like to merge this anyway. I probably just overlooked something while testing.
The new qudi core under development is implementing it a bit differently but maybe this PR can be ported to the new core as well.
Let's merge since it is not breaking anything, thank you.

PR #589, opened 2020-04-18
Add basic calculation tool of magnetic field with NV to Qudi. Also ad…

liuyanulm:
Added basic field calculation tool with NV center.
This PR is aimed to take place of PR #541 , basing on the discussions within it.

## Description
add a new logic and new Gui for some basic field calculation with NV center.

## How Has This Been Tested?
Tested on Qudi dummy.

## Types of changes

## Checklist:


NicolasStaudenmaier:
Hi Yan, I have tested your calculation tool and it seems to work well also on a running setup. But there are a few points which still need to be changed (partly as discussed in the meeting from Oct 7th [20201007_software_meeting](https://github.com/Ulm-IQO/software-meeting-documentation/blob/master/20201007_software_meeting.md)): 
1. The logic only takes fit results from a double Lorentzian fit. You should add also the possibility to use Gaussian fits. In general, there's not a very good survey of where the result comes from as the logic takes the fit results from another GUI. Personally, for me that's ok but maybe at least the frequency values which are taken should be displayed. 
1. As the connectors to the odmr and pulsed measurement logic are optional one should not be able to want to access them if they are not connected. Maybe you can query the connection when extracting fit values from either of the logics and raise an error message in case it is not connected. 
1. Similar for the frequency values which you want to extract from the fit. Give an error message if there are not the parameters which you want to give over and tell the user that a different fit has to be done first. 
1. The config option `data_source` gives a warning if it's not declared in the config. Please mark it as optional or use it in the example config (I would choose the first option as it's not really important at activation). 
1. On activation the function `set_data_source(self.data_source)` can be simply called. That way in the GUI also the correct data source will be displayed (if declared in config).

liuyanulm:
> Hi Yan, I have tested your calculation tool and it seems to work well also on a running setup. But there are a few points which still need to be changed (partly as discussed in the meeting from Oct 7th [20201007_software_meeting](https://github.com/Ulm-IQO/software-meeting-documentation/blob/master/20201007_software_meeting.md)):
> 
> 1. The logic only takes fit results from a double Lorentzian fit. You should add also the possibility to use Gaussian fits. In general, there's not a very good survey of where the result comes from as the logic takes the fit results from another GUI. Personally, for me that's ok but maybe at least the frequency values which are taken should be displayed.
> 2. As the connectors to the odmr and pulsed measurement logic are optional one should not be able to want to access them if they are not connected. Maybe you can query the connection when extracting fit values from either of the logics and raise an error message in case it is not connected.
> 3. Similar for the frequency values which you want to extract from the fit. Give an error message if there are not the parameters which you want to give over and tell the user that a different fit has to be done first.
> 4. The config option `data_source` gives a warning if it's not declared in the config. Please mark it as optional or use it in the example config (I would choose the first option as it's not really important at activation).
> 5. On activation the function `set_data_source(self.data_source)` can be simply called. That way in the GUI also the correct data source will be displayed (if declared in config).

Hi @NicolasStaudenmaier ,

Thanks again for the useful suggestions.

Please check out the modified version.

Now, it allows double gaussian fitting.

Some related comments. I removed config option for data source, because it looks actually not necessary. It should be similar way of how we select fitting functions in odmrlogic or pulsedmeasurementlogic.

liuyanulm:
it has been a very long time since this pr is created and revised according to @NicolasStaudenmaier suggestions.
I wish this pr can be merged into the master or some more comments on further revision!

NicolasStaudenmaier:
Hi Yan, I have reviewed and tested your module again. I suggest two more changes before we can merge. 
* When calculating the magnetic field strength and angle it is not tested whether the frequency input values are physically possible. This should be included and further calculation should be stopped and an error raised if the input is not physical. This means first of all checking if one frequency value is smaller and the other large than the zero-field splitting. Then it needs to be checked whether the angle and strength can actually be calculated, as there are further restrictions due to the `arccos` and the `sqrt` in the formulae. 
* When taking the frequency from `ODMR` or `pulsed measurement` and the current fitting routine is not correct, it would be nice to tell the user in the warning that the only two "allowed" fits are double Lorentzian and double Gaussian.

liuyanulm:
> Hi Yan, I have reviewed and tested your module again. I suggest two more changes before we can merge.
> 
> * When calculating the magnetic field strength and angle it is not tested whether the frequency input values are physically possible. This should be included and further calculation should be stopped and an error raised if the input is not physical. This means first of all checking if one frequency value is smaller and the other large than the zero-field splitting. Then it needs to be checked whether the angle and strength can actually be calculated, as there are further restrictions due to the `arccos` and the `sqrt` in the formulae.
> * When taking the frequency from `ODMR` or `pulsed measurement` and the current fitting routine is not correct, it would be nice to tell the user in the warning that the only two "allowed" fits are double Lorentzian and double Gaussian.


Hi Nicolas @NicolasStaudenmaier ,

Thanks for the testing!
About your new comments, here are the replies:

- The field calculation function actually is a physical filter itself. And the measured frequencies, more importantly, need to be verified by the user himself/herself.  Moreover, see attached picture (from https://static-content.springer.com/esm/art%3A10.1038%2Fnature07278/MediaObjects/41586_2008_BFnature07278_MOESM272_ESM.pdf ), when NV axis is perpendicular with field axis, both frequencies can be larger than the zero-field splitting. Therefore, I believe the current filtering setting is doing a good job.
- I revised the 'Tip' when user puts his/her mouse on the 'Auto Freqs' window, showing that only 'double Lorentzian or double Gaussian is applicable. what do you think?
- 
微信截图_20210414142402

liuyanulm:
Hi Nicolas @NicolasStaudenmaier ,

Unfortunately, I was removed from the 'qudi-commit team'. 
The github server is constantly denying my commit. 
It is like I was blacklisted by the qudi-iqo-administrator.
see attached images
微信截图_20210414144754

微信截图_20210414144814

liuyanulm:
Hi @Neverhorst , Hi @ksenkalla,

Did you accidently (or on purpose :(   ) remove me from the 'commit team'? As a consequence, I cannot commit any more, see the conversation between me and Nikolas.

NicolasStaudenmaier:
> Hi Nicolas @NicolasStaudenmaier ,
> 
> Thanks for the testing!
> About your new comments, here are the replies:
> 
> * The field calculation function actually is a physical filter itself. And the measured frequencies, more importantly, need to be verified by the user himself/herself.  Moreover, see attached picture (from https://static-content.springer.com/esm/art%3A10.1038%2Fnature07278/MediaObjects/41586_2008_BFnature07278_MOESM272_ESM.pdf ), when NV axis is perpendicular with field axis, both frequencies can be larger than the zero-field splitting. Therefore, I believe the current filtering setting is doing a good job.
> * I revised the 'Tip' when user puts his/her mouse on the 'Auto Freqs' window, showing that only 'double Lorentzian or double Gaussian is applicable. what do you think?
> 
> 微信截图_20210414142402

Hi @liuyanulm, 
I agree that the user should know about the frequency values that he/she inserts. I also see that both frequency values can be larger than the ZFS, but nonetheless there are values that lead to unphysical behavior. In the calculation the result is then `np.nan`. As also typos can easily occur the user should be informed if the frequency values cannot be correct. This is not much work and it just needs to be checked after the calculation (l.130) `if np.isnan(angle) or np.isnan(b_field):`. In that case raise an error with `self.log.error()` which says that the frequency values are physically not possible. 

About the second point. Actually I haven't noticed yet that there is this description. In any case I think it is more useful for the user to tell him in the log if something with the fit is not correct. As there is already the warning message in line 155 please just add that the current fit method is not suitable and it has to be either double Lorentzian or double Gaussian. 

Please quickly change these two points as I believe it would make tracing of misbehavior much easier. Then we can finally merge this nice tool.

liuyanulm:
> > Hi Nicolas @NicolasStaudenmaier ,
> > Thanks for the testing!
> > About your new comments, here are the replies:
> > 
> > * The field calculation function actually is a physical filter itself. And the measured frequencies, more importantly, need to be verified by the user himself/herself.  Moreover, see attached picture (from https://static-content.springer.com/esm/art%3A10.1038%2Fnature07278/MediaObjects/41586_2008_BFnature07278_MOESM272_ESM.pdf ), when NV axis is perpendicular with field axis, both frequencies can be larger than the zero-field splitting. Therefore, I believe the current filtering setting is doing a good job.
> > * I revised the 'Tip' when user puts his/her mouse on the 'Auto Freqs' window, showing that only 'double Lorentzian or double Gaussian is applicable. what do you think?
> > 
> > 微信截图_20210414142402
> 
> Hi @liuyanulm,
> I agree that the user should know about the frequency values that he/she inserts. I also see that both frequency values can be larger than the ZFS, but nonetheless there are values that lead to unphysical behavior. In the calculation the result is then `np.nan`. As also typos can easily occur the user should be informed if the frequency values cannot be correct. This is not much work and it just needs to be checked after the calculation (l.130) `if np.isnan(angle) or np.isnan(b_field):`. In that case raise an error with `self.log.error()` which says that the frequency values are physically not possible.
> 
> About the second point. Actually I haven't noticed yet that there is this description. In any case I think it is more useful for the user to tell him in the log if something with the fit is not correct. As there is already the warning message in line 155 please just add that the current fit method is not suitable and it has to be either double Lorentzian or double Gaussian.
> 
> Please quickly change these two points as I believe it would make tracing of misbehavior much easier. Then we can finally merge this nice tool.

Hi Nicolas @NicolasStaudenmaier ,

Thanks for the detailed suggestions!
I will do the revision, and meanwhile I hope that the administrators can un-blacklist me, so I can push the commit to this branch.

ksenkalla:
> Hi @Neverhorst , Hi @ksenkalla,
> 
> Did you accidently (or on purpose :( ) remove me from the 'commit team'? As a consequence, I cannot commit any more, see the conversation between me and Nikolas.

Hello @liuyanulm , 
thanks for your ongoing work on this. I give you access back to the qudi-commit team until this PR is finished. Then it is possible to continue collaboration as outside collaborator. The process to commit there is quite similar.

liuyanulm:
> > Hi @Neverhorst , Hi @ksenkalla,
> > Did you accidently (or on purpose :( ) remove me from the 'commit team'? As a consequence, I cannot commit any more, see the conversation between me and Nikolas.
> 
> Hello @liuyanulm ,
> thanks for your ongoing work on this. I give you access back to the qudi-commit team until this PR is finished. Then it is possible to continue collaboration as outside collaborator. The process to commit there is quite similar.

Hi @ksenkalla , @Neverhorst ,

Thanks for putting me back to the list. I have done some update of the branch.

However, it looks from my side that, I cannot commit or push changes to qudi on GitHub if I got removed from the list. I wonder removing me again would probably bans me from further commits. I am pretty doubful if I could commit like a "outside collaborator".

Additionally, on the same list, I notice other former members of IQO staying in the list.

NicolasStaudenmaier (approved):
Thanks Yan! Looks good to me now and can be merged.

PR #594, opened 2020-05-06
improve loading StatusVar for dicts

LindnerIAF:
Optimizing the loading of StatusVars

## Description
If the Status Vars tries to load a dict type item, it will now load the default values for every key, if the key is missing, and not only if the entire dictionary is missing.

## Motivation and Context
If someone adds a key to an exisiting dict, this key with its value will only be loaded, if the StatusVar is newly loaded. One example would be adding a channel to the channel list at the pulsed logic. If I add a new channel for a new trigger output, the corresponding channel option will only appear after I delete the StatusVar save file or the entry.

## How Has This Been Tested?
tested on a confocal setup and dummy. 

## Types of changes

## Checklist:

### DIFF
--- a/core/module.py
+++ b/core/module.py
@@ -205,7 +205,11 @@ class BaseMixin(metaclass=ModuleMeta):
         # add status vars
         for vname, var in self._stat_vars.items():
             sv = self._statusVariables
-            svar = sv[var.name] if var.name in sv else var.default
+            if isinstance(var.default, (OrderedDict, dict)) and var.name in sv:
+                svar = var.default
+                svar.update(sv[var.name])
+            else:
+                svar = sv[var.name] if var.name in sv else var.default
 
             if var.constructor_function is None:
                 setattr(self, var.var_name, svar)
--- a/core/module.py
+++ b/core/module.py
@@ -19,6 +19,7 @@ Copyright (c) the Qudi Developers. See the COPYRIGHT.txt file at the
 top-level directory of this distribution and at <https://github.com/Ulm-IQO/qudi/>
 """
 
+import copy
 import logging
 import warnings
 from fysom import Fysom  # provides a final state machine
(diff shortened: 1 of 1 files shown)


kay-jahnke:
@Neverhorst can you take that over to the core as well.

core/module.py
```diff
@@ -205,7 +205,11 @@ def __load_status_vars_activate(self, event):
         # add status vars
         for vname, var in self._stat_vars.items():
             sv = self._statusVariables
-            svar = sv[var.name] if var.name in sv else var.default
+            if isinstance(var.default, (OrderedDict, dict)) and var.name in sv:
```
Neverhorst, line 208: Can be simplified by solely checking for `dict` (`OrderedDict` is subclass of `dict`)

core/module.py
```diff
@@ -205,7 +205,11 @@ def __load_status_vars_activate(self, event):
...
+                svar = var.default
+                svar.update(sv[var.name])
+            else:
+                svar = sv[var.name] if var.name in sv else var.default
```
Neverhorst, line 212: Just realized that whenever `var.default` is used to initialize the `StatusVar`, a deepcopy of this default value should be assigned. Otherwise the module will alter the default value of the `StatusVar` instance which could lead to undefined behaviour. So please exchange both `var.default` occurrences with `copy.deepcopy(var.default)`.
This has been an oversight before, so let's fix it along with this PR.

LindnerIAF: Thank you for the review. I add this. Theoretical no module read the default value beacause there is a copy in the mata class , but just in case.

Neverhorst: Well, there is a deepcopy created in the metaclass. However, if you just deactivate and activate the module again, the exact same copy is being used because `ModuleMeta.__new__()` is not being called again. Thus you can construct problematic edge cases from this. And the `StatusVar` default value was definetely the same identity as the exposed value in the module (`var.default is my_module.var_name` evaluates `True`).
Of course under "normal" operation this is not causing any problems but let's improve it anyway.
It is very likely that the deepcopy in `ModuleMeta` can be omitted instead, but since there is a new revision of the entire qudi core underway that changes the module base class and loading procedure quite substantially, I think it's not worth investigating. 
Thank you again.

Issue #552, opened 2019-10-17
Multi interface / Fastcounter problem


alrik-durand:
## *What* is affected by this bug?
We are trying to use the new multi interface feature  (#542) to use a hardware module as slow and fast counter. We have a working version, but for some mysterious reason, the activation fails about one every two time.

## *When* does this occur?
When we try to load the Pulsemeasurement GUI, and it tries to load all the sub modules. Sometime we activate this one first, sometime we activate the slow counter GUI before, there is no obvious correlation.

## *Where* on the platform does it happen?
We have observed this on Windows 10 with last version of Qudi

## *How* do we replicate the issue?
I created a branch called "timeharp260_dummy" on my repo with the basic code necessary to replicate the issue. I removed all the dll call so that it can be replicated on a computer without the physical hardware.
I also modified the default.cfg so that it is easy to test. Just load Qudi on this branch, then start pulsemeasurement GUI, and one every few time you get the error given bellow.

## Other Comments

The error given by Qudi is : 

`
Error during activation
KeyError: 'No method registered for interface "FastCounterInterface".'
Traceback (most recent call last):

  File "C:\Users\alrik\Documents\GitHub\qudi\core\module.py", line 96, in wrap_event
    base_event(*args, **kwargs)

  File "C:\Users\alrik\AppData\Local\Continuum\anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 311, in fn
    self.transition()

  File "C:\Users\alrik\AppData\Local\Continuum\anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 306, in _tran
    self._after_event(e)

  File "C:\Users\alrik\AppData\Local\Continuum\anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 338, in _after_event
    return getattr(self, fnname)(e)

  File "C:\Users\alrik\AppData\Local\Continuum\anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 95, in _callback
    return func(obj, *args, **kwargs)

  File "C:\Users\alrik\Documents\GitHub\qudi\core\module.py", line 216, in __load_status_vars_activate
    self.on_activate()

  File "C:\Users\alrik\Documents\GitHub\qudi\gui\pulsed\pulsed_maingui.py", line 191, in on_activate
    self._activate_analysis_ui()

[...]

PR #601, opened 2020-06-02
Multi interface fix

alrik-durand:
## Description
I've corrected a bug in the class InterfaceMethod for overloading hardware methods for multiple interfaces.

Once a module using the register for get_constraints was loaded, any new module declaring a get_constraints method would mess up the first one. This is because every InterfaceMethod share a common _latest_unregistered_instances dictionary.
When cleaning it,  func.__name__ was the name of the interface specific method (ex : some_arbitrary_name1) instead of the name of the overloaded method (my_overloaded_method).

## Motivation and Context
This solve an active bug making the overloading of interfaces quite difficult.  The bug was reported in issue #552 .
The bug is hard to track as it might depend on the loading order of the modules. For example, loading pulsed_measurement_logic activate both the fastcounter and the pulser in a non deterministic order.

## How Has This Been Tested?
I've tested this on a dummy file implementing SlowCounter and FastCounter. 

## Types of changes

## Checklist:

### DIFF
--- a/core/interface.py
+++ b/core/interface.py
@@ -48,6 +48,7 @@ class InterfaceMethod:
 
     def __init__(self, default_callable, is_abstractmethod=False):
         self._obj = None
+        self._name = default_callable.__name__
         if is_abstractmethod:
             self._default_callable = abstractmethod(default_callable)
         else:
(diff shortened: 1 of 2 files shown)


alrik-durand:
Thanks for the response.

I'm not sure of the why, but during the debugging, when loading pulsemeasurementlogic as first module, the error was not raised every time (about 2 every 3 time on my PC).
The only explanation that came to my mind was parallel loading !

PR #603, opened 2020-06-18
Thorlabs flipper - Switch interface

alrik-durand:
## Description
I've added a hardware module to interface Thorlabs flipper MF101 with the switch interface.
It uses the recommended method via Kinesis dll.

## Motivation and Context
This PR add support to new material.

## How Has This Been Tested?
This has been tested with a MFF101/M on windows 10.

## Types of changes

## Checklist:


alrik-durand:
I've updated the hardware to conform to the new interface.
I had to copy/past the *states* functions from the interface because otherwise I get an error. There seems to be a conflict between interface overloading (not used by this hardware) and attribute/setter methods.
The copying and pasting of the default implementation prevent the problem, but it should probably be solved properly.

Neverhorst:
@alrik-durand: See PR #641

Neverhorst:
Did PR #641 resolve the issue mentioned in your latest comment, @alrik-durand?
Should not be necessary to copy the `states` method over anymore.

Issue #593, opened 2020-04-29
Scientific Spinbox and Mapper are incompatible


tobiasgehring:
The mapper expects the widget to have a Qt Property for the value. However, scientific spinbox does not define `value` as a property as regular spinboxes do.

I'm not sure if it is possible to define `value` as Qt property in python as the getter has the same name. It's apparently a standard in Qt that they share the same name, but a quick look into the documentation did not reveal if this is possible within python. Probably not.

2 possible solutions:

1) add an attribute to the scientific spinbox, e.g. called `value_int` (or something else) which is the Qt property. This is a one line change per spinbox. This property could then be used with mapper. Naming is non-standard though as the property of a regular spinbox is called `value`.

2) Change mapper to also accept getter and setter instead of a Qt property and use this if mapper is used with the scientific spinbox.

My quick fix is 1, but I don't want to make a pull request without comments.

Neverhorst:
Well, my preferred solution would be to revise the scientific spinboxes to make use of `QtCore.Property` for all properties. This should not only solve the mapper issue but also get rid of some minor bugs when using these widgets in QtDesigner.

After a quick read on `QtCore.Property` in PyQt5 it seems that you can not use them as classic Python property which causes trouble for users/programmers not familiar with Qt.
However since we are moving to PySide2 this approach is becoming more appealing. There they state that a `QtCore.Property` functions as Qt AND Python property. (https://wiki.qt.io/Qt_for_Python_UsingQtProperties)

So depending on how fast you need it, I would suggest revising the scientific spinbox in the upcoming PySide2 qudi core and optionally fix it for now in the least invasive way possible (probably your suggestion no. 1)

See also: PR #605

Neverhorst:
Fixed with PR #605

PR #605, opened 2020-06-22
Added ScienDSpinbox compatibility to mapper

alrik-durand:
## Description
I have added the compatibility of SiencDSpinbox and mapper.

Also this PR brings the documentation update from PR #587 which is now a bit obsolete.

## Motivation and Context
Mapper module helps develop GUI modules by taking care of the repetitive task of mapping the state of the model to the state of the view in the GUI. It is still not used very much.
ScienDSpinbox it an layer on top of Qt Spinbox to help deals with Scientific unit. It is used in most of the spinbox of Qudi. 
Compatibility between the two was missing, this small PR adds such support.

## How Has This Been Tested?
This PR has been tested with code from a PID overhaul that will be included in another PR.

## Types of changes

## Checklist:

### DIFF
--- a/core/mapper.py
+++ b/core/mapper.py
@@ -31,6 +31,8 @@ from qtpy.QtWidgets import QLineEdit
 from qtpy.QtWidgets import QPlainTextEdit
 from qtpy.QtWidgets import QSpinBox
 
+from qtwidgets.scientific_spinbox import ScienDSpinBox
+
 import functools
 
 SUBMIT_POLICY_AUTO = 0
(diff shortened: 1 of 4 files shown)


tobiasgehring:
This is one option. Another option, which may only require minimal changes, is to edit the spinbox itself and add
`value_float = QtCore.Property(float, fget=value, fset=setValue, notify=valueChanged)`
With this qt property we just need to add two lines around line 130 in mapper to return `value_float` if instance is the scientific spinbox.
I think this solution would be more elegant.

tobiasgehring:
see #593

alrik-durand:
I agree with both of you that Option 1 is cleaner. Thank you.
I'm not very familiar with the GUI side of Qudi so I sometimes try unnecessary complicated options !

I implemented the changes mentioned and tested it on a branch using mapper ad ScienDSpinbox for a pid_overlaul, it works perfectly.

Should I change anything else ?

Neverhorst:
Um... should the mapper at line 130 not return `value_float` (instead of `value`)?

alrik-durand:
You are totally right... I tested it while specifying the *widget_property_name* the GUI but this is obviously better. Fixed.

Neverhorst:
OK cool. Would you also directly fix it for the `ScienSpinBox` (int) as well? Should be the same process with `value_int` instead.

alrik-durand:
No problem ! I've changed it.

Neverhorst (approved):

PR #607, opened 2020-07-01
Thorlabs filter wheel FW102C

alrik-durand:
## Description
This PR adds support of the FW102C filter wheels from Thorlabs as a hardware module.

## Motivation and Context
Thorlabs filter wheels are generally interfaced by their dedicated software. For some application, it's necessary to operate them from a Qudi script. This PR adds this possibility.
As there is no appropriate interface for this hardware, it does not have any.  This means this module is only compatible with scripts for now.

## How Has This Been Tested?
This has been tested with a FW102C and a six position wheel on a windows 10 PC.

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -77,6 +77,7 @@ please use _ni_x_series_in_streamer.py_ as hardware module.
 * Set proper minimum wavelength value in constraints of Tektronix AWG7k series HW module
 * Added a hardware file for fibered optical switch Thorlabs OSW12/22 via SwitchInterface
 * Fixed bug affecting interface overloading of Qudi modules
+* Added a hardware file to interface Thorlabs filter wheels via scripts
 *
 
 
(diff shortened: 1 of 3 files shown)


Neverhorst (changes requested):
I have the following remarks:

- Please give the module and the class a more descriptive name (instead of "thorlabs" and "Main").
- It is not a good idea to store the PyVISA resource manager and device instance as class attribute. Imagine controlling 2 or more filter wheels in the same qudi process, resulting in multiple instances of `Main`. In fact the class attributes are never used anyway since assigning `self._rm = ...` and `self._inst = ...` will create an instance attribute with the same name. So just remove the two attributes from the class body please.

alrik-durand:
Thank you for your feedback, 
I'll stop giving "main" as class name then !
You are totally right about this second point, I fixed this.

Neverhorst (approved):
Thank you

PR #609, opened 2020-08-17
Added scripting helper methods

alrik-durand:
## Description
This PR adds general helper methods to multiple logic modules : time series, poi manager and odmr logic.
These are non breaking change.
The helper at the end of the logic files are meant to be called by scripts or tasks within their own threads.

## Motivation and Context
These changes simplify the use of Jupyter notebook (or tasks) by defining helper methods that are not used by the module itself at the end of the file.
This is something that was already being done in Qudi like for example `perform_odmr_measurement` of *odmr_logic*. 

## How Has This Been Tested?
All the methods have been tested with dummy hardware and behave as expected.

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -77,6 +77,7 @@ please use _ni_x_series_in_streamer.py_ as hardware module.
 * Set proper minimum wavelength value in constraints of Tektronix AWG7k series HW module
 * Added a hardware file for fibered optical switch Thorlabs OSW12/22 via SwitchInterface
 * Fixed bug affecting interface overloading of Qudi modules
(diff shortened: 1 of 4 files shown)


Neverhorst:
I think adding scripting methods to the logic modules is not a good coding practice. It bloats the already big logic modules even more and does not contribute to the respective logic "API". In other words the logic should not contain its own wrapper. 
Adding small scripts to facilitate scripted experiments sparked the idea of tasks in qudi. However the task runner and automation tool are non-functional and will be removed in a future release.
But there will be a similar mechanic in the new core release to use scripting objects inspired by the old task logic.

PR #610, opened 2020-08-25
Core bug fixes

kay-jahnke:
- Making Error messages sticky: The Error messages now stay on top but they do not need to be focused (non-modal).
- deleting the restart method in the manager and rather making use of the realQuit, because only this one respects the dependencies

## Description

When clicking the restart Button in the manager qudi was restarted, but the individual modules were deloaded in "random" order. This lead to cases where the hardware module was deactivated while the logic still tried to access it. The realQuit does respect the dependencies, so it was easy to just extend it to also handle the restart. the old restart method was then deleted.

Also ran an automatic code cleaning on the files I touched.

## Motivation and Context

- Error messages got lost when too many windows were open on the computer.
- Restarts of qudi produced unnecessary errors that might also damage hardware

## How Has This Been Tested?

tested with dummy config(s) on Linux 5.7.9-1-MANJARO

## Types of changes

## Checklist:

### DIFF
--- a/core/manager.py
+++ b/core/manager.py
@@ -1372,6 +1372,7 @@ class Manager(QtCore.QObject):
             self.realQuit()
 
     @QtCore.Slot()
+    @QtCore.Slot(bool)
     def realQuit(self, restart=False):
         """ Stop all modules, no questions asked. """
         deps = self.getAllRecursiveModuleDependencies(self.tree['loaded'])
--- a/core/manager.py
+++ b/core/manager.py
@@ -1149,30 +1149,6 @@ class Manager(QtCore.QObject):
             If the module is an active GUI module, show its window.
         """
 
-        print(base,
-              key,
-              'not equal' if QtCore.QThread.currentThread() is not self.thread() else 'equal',
-              QtCore.QThread.currentThread(),
-              self.thread())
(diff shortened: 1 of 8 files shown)


core/manager.py
```diff
@@ -1154,6 +1157,17 @@ def startModule(self, base, key):
             If the module is already loaded, just activate it.
             If the module is an active GUI module, show its window.
         """
+
+        if QtCore.QThread.currentThread() is not self.thread():
+            QtCore.QMetaObject.invokeMethod(
```
Neverhorst, line 1162: Might this cause the trouble I mentioned?

Neverhorst (commented):
While testing with default config, qudi hangs after opening the manager window (probably while automatically loading modules for tasklogic) and does not respond anymore. I suspect the lines 1160ff in _manager.py_.

core/manager.py
kay-jahnke: Thanks for the tests, I think I will need to double check on Windows. I will then come back to you.

kay-jahnke: I deleted this part of the pull request because it is not easily solvable in the old core. New core handles it automatically.

Neverhorst (approved):
Seems to work now. Shall we merge?

PR #612, opened 2020-08-28
adding Keysight AWG M8190A

timoML:
- new: hardware file for keysight awg
- added: pulser constraint to indicate linear sequence table
- normal pulsed measurements: stable
- sequence mode: may containt smaller bugs, but works in general

## Motivation and Context
Introduces new device to qudi that is increasingly used at IQO. Note that this is NOT the hardware file for the keysight M8195A, which is more widespread. Merging these two into one file might be a future project.

## How Has This Been Tested?
On setup3:
- has been in use on own branch for ~ 2 months incl. sequence mode
- within the current master:
- started qudi and loaded the pulsedmeasurementgui
- uploaded vaious predefined waveforms 
- tested pulsed measurments with results as expected in laser_on, podmr, rabi

## Types of changes

## Checklist:

### DIFF
--- a/documentation/changelog.md
+++ b/documentation/changelog.md
@@ -223,6 +223,7 @@ This can be used to specify the axis labels for the measurement (excluding units
     * Introduced separate fit tools for each of the two plots in the pulsed analysis tab
     * Automatically clears fit data when changing the alternative plot type or starting a new 
       measurement.
+    * Added Keysight AWG8190A hardware file.
 
 Config changes:
 * **All** pulsed related logic module paths need to be changed because they have been moved in the logic
--- a/hardware/awg/keysight_M8190A.py
+++ b/hardware/awg/keysight_M8190A.py
@@ -281,7 +281,9 @@ class AWGM8190A(Base, PulserInterface):
         constraints.sequence_num.default = 1
 
         constraints.sequence_option = SequenceOption.OPTIONAL
-        constraints.sequence_order = SequenceOrderOption.LINONLY
+        # temporary property to check from outside
+        # not available in pulser interface or any other awg
+        constraints.sequence_order = "LINONLY"  #SequenceOrderOption.LINONLY
 
         # If sequencer mode is available then these should be specified
         constraints.repetitions.min = 0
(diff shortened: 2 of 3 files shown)


Neverhorst (changes requested):
Please exclude the change to _pulser_interface.py_ since this new constraint is not being used by any logic module. Also if you introduce a change to the interface, you would need to adapt all available hardware modules to this change.

Please drop a line about the newly added hardware in th changelog.

The docstring of the new hardware module class needs to include a complete config example to copy-paste into the config file. (see other modules)

timoML:
As there is currently no general way to implement a sequence table using <go_to> statements on devices with a linear sequence table, I would vote to keep the SequenceOrderOption in the interface. Reasoning:
- The type of supported sequence table of the device is "on the same level" as the already existing SequenceOption.
- Since the PulserConstraints is default init to SequenceOrderOption.ARB, it is non-breaking with existing devices.
- Sequences created in some logic module (eg via a predefined method) need a way to check whether <go_to> is available on the pulser.

If you insisted on keeping the interface untouched, I see only two alternatives here:
- Wait until a general algorithm is contributed that replaces <go_to> for devices with linear sequence tables.
Even then, this logic would require to check the type of pulser via some flag.
- Implement a property in devices with linear sequence tables and do some try/except checking on the property in logic.

Happy to hear better ideas... :)

Changes to changelog and docstring as requested.

Neverhorst:
2 options to merge this PR:

1. keep the interface change
**consequence:** Add the constraint to all pulser hardware modules in qudi.

2. ditch the interface change _for now_
**consequence:** Nothing immediate

In both cases I agree we need to check for something like this in `sequence_generator_logic.py` in the future. I do not question the reasoning for that. But this is not implemented yet, thus one of the options above. 
As long as the logic does not account for this new constraint, `write_sequence()` in keysight hw-module needs to check for consecutive sequence steps in both cases and indicate error if this is not fulfilled (by returning `-1` and maybe also logging an error to inform the user why it has failed).

timoML:
Should be ready to merge now.

Neverhorst:
@timoML: Can this PR be closed in favor of the new PR #634 ?

PR #620, opened 2020-10-09
Minicircuit SPDT switch hardware

alrik-durand:
## Description
I've added a hardware file to support minicircuit SPDT hardware via SwitchInterface

## Motivation and Context

This hardware controls one or multiple switch via SMA cables. It can connect either port 1 or port 2 to a COM port.
 This type of hardware can automatize the change of cabling configuration for SMA, BNC, etc. cables.
This PR adds compatibility to it via switch interface.

## How Has This Been Tested?
This has been tested with a brand new RC-4SPDT-A26 on the local network and a windows 10 computer.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [x] I have documented my changes in the changelog (`documentation/changelog.md`)
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added/updated for the module the config example in the docstring of the class accordingly.
- [x] I have checked that the change does not contain obvious errors (syntax, indentation, mutable default values).
- [ ] I have tested my changes using 'Load all modules' on the default dummy configuration with my changes included.
- [x] All changed Jupyter notebooks have been stripped of their output cells.

Issue #586, opened 2020-04-09
[Idea] Architecture Decision Record


alrik-durand:
Should Qudi have a Architecture Decision Record (ARD) ? 

> An architectural decision record (ADR) is a document that captures an important architectural decision made along with its context and consequences.

Let me explain,
We are currently working on a new set of modules in our lab, and we were discussing on the architecture decisions for the interface. One example of such question is "what should a setter return ?". We tried to look at other interfaces for help, but they are quite heterogeneous.

Using an ARD could help with this problem, by creating a place where a joint decision is recorded and explained, in the context of the Qudi software.

Here is a link with more detail about what ARD are and how they can me implemented 
https://github.com/joelparkerhenderson/architecture_decision_record

What do you think of the proposed idea ?

Alrik

PR #623, opened 2020-10-22
Switch interface rework

kay-jahnke:
The whole tool chains for simple switches has been reworked. The hardware controlled by this can simply be switched between any two or more different states without the need for any timing (software timed).

**Attention:** This is a change to an interface, so it is a breaking change for old switch hardware wrappers not included in qudi that use the SwitchInterface!

## Description

The old tool chain was quite old and did not conform to the qudi guidelines by using camel-case, old connector types and not making use of signal. The new logic now conforms to the guidelines and make excessive use of pythonic properties and setters.

- A change of the switch state in the logic in now reflected in the gui by transmitting signals
- Names can be given to states, switches and the hardware itself
- Switches are referenced by their given name
- States are be referenced their name

The GUI has changed to Radio Buttons to reflect the option to set more than 2 states:
Coloured radio buttons are optimized for better readability when wearing laser protection goggles in the lab:
image

## How Has This Been Tested?

Tested on Linux 5.8.11-1-MANJARO with dummy config.
Please additionally test with the hardware when you have it available.

## Types of changes

## Checklist:


kay-jahnke:
@alrik-durand I think the Thorlabs ows12 module was originally from you. As I don't have any of these, could you please test it with your hardware.

Also it would be perfect if you adapted PR #620 to the new interface, once this PR is through.

alrik-durand:
Hi Kay, thank you for this rework, this is a big improvement !

I will test the OSW12 asap ! Also my PR #603 will need rework, I'll rework both PR as soon as I can.

Can I make a suggestion about switch logic ? Some devices have a physical button to change the state of the switch. For them the value can change without the logic being aware of. What do you think about having an optional (via config option) loop that check periodically (also via config) the state of the hardware ?

kay-jahnke:
> Can I make a suggestion about switch logic ? Some devices have a physical button to change the state of the switch. For them the value can change without the logic being aware of. What do you think about having an optional (via config option) loop that check periodically (also via config) the state of the hardware ?

OK, give me a minute.

kay-jahnke:
@alrik-durand here you go.

More ideas?

alrik-durand:
Thanks, I'm not sure it's useful for everybody though, maybe the watchdog can be optional ? 

I'm trying osw12 hardware right now, there is some bugs but I'll try to fix them !

I have some opinions on some more stuff, I'll give you a long version once I finish this.

Neverhorst:
Long overdue rework... thank you very much.

I know we discussed that already briefly, but if you would use state dicts instead of indices you would not need indices AND names (less complexity and more "pythonic"... see [ZEN of Python](https://www.python.org/dev/peps/pep-0020/), especially the parts about: _There should be one-- and preferably only one --obvious way to do it_ and _Explicit is better than implicit_).
You would also be able to set multiple but not all switches with a single function call.
And as a bonus, since `dict` is by default ordered since Python 3.6, you can STILL select switches by index if you absolutely must.
But if you already require names for each switch, it is much more clean to consequently address everything by name. (this is Python, not C after all... :sweat_smile:)

I would propose state dicts to look somehow like this when getting the states:
```Python
state_dict = {'my_switch'      : (True, 'state_on_name'),
              'my_other_switch': (False, 'state_off_name')}
```
For setting them you could use either `bool` or `str` instead of the tuple.

Neverhorst:
Another thing that comes to my mind:
Would it not make sense to lay out the interface to use multiple state switches instead of binary switches (`True/False`).
It should be a minor addition since instead of `bool` you could just use a state `Enum` and specify valid states in the hardware constraints.

kay-jahnke:
@Neverhorst what do you think of this: currently just implemented for switch_dummy.py without interfuse.

alrik-durand:
I managed to get the osw12 working (I've pushed on switch_interface_rework) on my fork. Weirdly there are still some problem because the hardware does not respond with the latest update all the time for some reason. This is something I can live with.

I like Neverhorst's idea of having more that two states, I know some usage that could benefit from this in our lab.

I have a few remarks concerning the current implementation of the new switch rework.

1. Right now each hardware module has to implement a lot of "advanced" stuff so that the GUI can be nice and user friendly. I'm open to other opinion about this but in my understanding, the logic module should do as much "general stuff" as possible and let the hardware only worry about the implementation specific features specified by the interface. 

For example, it look like the code used to give name to each switch and switch_state is the same in every hardware. This gives a nice UI, but maybe the same can be accomplished by doing this in the logic ?

2. In this version some modules have a Status variable to remember the state of the switch. I'm not a big fan of this, I don't like Qudi to change things unnecessarily when starting up. I would prefer Qudi to just read the current switch status at startup and keep it, without the need for status variable. If you need this anyway, maybe make it optional, and again in logic as every hardware could use it

3. Current logic does not hold a "model" of the current state. This means that each time the logic needs the switch state value, it ask the hardware for it. Again I would gladly discuss this (I even proposed #586 for this kind of things), but I think the logic module should hold the current state of the model. This has a few advantages, like reducing unnecessary serial communication or being more resilient to hardware problems.

4. One small thing : in your interface, the *set_state* setter doc says the function should return the new state value. This is something that is (sometimes) done in a few existing Qudi interfaces, but never really used. For example your logic do not use returned value. I don't think a setter should return something : if the logic wants the new version, the logic can get it explicitly, if the logic want to check for an error, python error system is more pythonic than error code.

If you have any question don't hesitate, I'm happy to elaborate.

kay-jahnke:
Hi @alrik-durand 
as you might have guessed from the changes in the code, I had some discussions with @Neverhorst in the background. My full fledged flexibility was putting to much overhead into the hardware files and causing some unexpected behaviour. 
This is why we now switched to named switches and states only: There should be one clear way to access a hardware file and the property should return something similar to what is input in the setter. Such functionality should live in logic, if at all. Also two different set_ and get_ methods were not sensible, therefore I deleted one.

Coming to your suggestions:
1. All the advanced stuff is mostly related to the naming of states, switches and hardware. By now this is independent of the GUI. It really depends on the hardware you want to control. Therefore, I'm strongly advocating of leaving it where it is.
Nevertheless, not all hardware needs the full flexibility, so the names do not have to have ConfigOptions in every hardware file but they can just be defined static.
2. The StatusVar actually started out in the NI Card and the dummy, because in both cases you can't just read the current switch state from the hardware. But you are correct, this is not needed for every hardware. There is actually a ConfigOption that should have made this behaviour optional (reset_hardware) but I implemented it incorrectly. I will change that to an optional "remember_state" so people can use it if they want, but don't need to.
3. I have a strong opinion on having different states between logic and hardware (this is what your proposal would end up being): This is a big NONO! As the calls between hardware and logic module are not synchronized, this would lead to hardware states being shown different than they are. If you actually want the behaviour you described, you should write an interfuse. I had to do that once, because we had an very annoying IC that did not like to be queried several times a second. With the interfuse you can run a timer that queries the values periodically and hands them to the logic when requested. The interfuse has the advantage that it makes this explicit. The other argument is, that the query behaviour is hardware specific and should therefore be handled by the hardware file and not the logic. Otherwise all connected hardware would inherit this annoying behaviour.
4. I complete agree. This now became obsolete, as I deleted the get_ and set_ methods.

I will now implement the new scheme for all hardware files (I will take over your bug fix) and would like you to review again afterwards.

Neverhorst:
I just want to elaborate a bit on your point 4, @alrik-durand, regarding the return values of setter methods because it may shed some light on design decisions of the past.

In a pure Python application I would completely agree with you that return values should not be used to check for success.
The "pythonic" way would be to just raise and always propagate exceptions up until they matter and must be handled by
 `try ... except`.
The problem with qudi is that this principle can not be rigorously applied between modules because most inter-module communication is handled via `Qt` signals. In this case "pythonic" error propagation does not work anymore.
So we have to violate the Python Zen just far enough to work with `Qt` which comes from a `C++` world.
This is exactly why we implemented our own error reporting system in order to not interrupt the rather complex signalling network flow by python exceptions.
Simply put, this is why we do not simply raise exceptions but call e.g. `self.log.error()` or `self.log.exception()` which does not break code execution. 

Coming back to return values...
The only inter-module communication where you could actually catch exceptions in a "pythonic" way, is between logic and hardware modules. These interfaces tend to be called directly by the logic (not via `Qt` signal).
However, this is not enforced by design since we wanted to keep the possibility to call hardware module methods via signal if neccessary (hardware modules can be forced to run in their own thread after all).
So the easiest way for the calling module to register a failing task and clean up after itself is to get a return value carrying an error code (or emit a signal that causes the calling module to clean up... ).

An entirely different story are hardware setter methods that return the actually set values.
This has nothing to do with error checking (although it could be used as such) but rather to account for `float` rounding, truncation, entanglement of dependent variables or any other slight alteration of the values to set performed by the actual hardware itself. These return values will be propagated back to the master module (mostly GUI) to inform the user about the REAL setting. So one does not generall assume the value to set to be perfectly represented by the hardware.
So I do not see any other generally applicable way to account for imperfectly set hardware parameters.

And by the way... there are a number of logic modules that account for hardware return values, e.g. [TimeSeriesReaderLogic](https://github.com/Ulm-IQO/qudi/blob/master/logic/time_series_reader_logic.py) and [PulsedMeasurementLogic](https://github.com/Ulm-IQO/qudi/blob/master/logic/pulsed/pulsed_measurement_logic.py), although not the majority...

If you have any better idea on how we can ensure the afore mentioned goals while being more "pythonic", please let us know.

All this concluded the current "rules" for qudi:
- Hardware setter methods should return the _actually_ set values
- Hardware methods that perform some task should return an error code or a failure indicator
- Logic should check for hardware return values
- Do not interrupt the code execution and practice defensive programming in order to notify the user of as many exceptions as possible through the built-in qudi logging facility (do not silently ignore what should be an exception raise in classic Python)
- Module methods that are intended to be called from outside the module (public methods) must not raise
- Hardware setters should always be exposed as classic method in order to be used as `Qt` slot. Python property setters must be just complementary if desired.

alrik-durand:
@kay-jahnke 

About point 3, I'm not sure we understand each other : 
I'm not saying the hardware module should keep the state of anything, I advocate for it to hold nothing (The only exception being for the rare instruments where some parameters that can be set but not read). Logic modules on the other hand needs to keep track of every states, for example _pulsed_measurement_logic_ keep track of most parameters (like __fast_counter_binwidth for example). I think the logic module should not query the hardware module every time the value is needed. For example, pulsed_measurement does this for one things : _is_gated_, this caused me some issues one day when the computer running the hardware module crashed and I could not save the data on the other computer running the logic because of this (I wrote a possible fix for this but did not PR this yet)
This approach assume the value can not be changed behind qudi's back, which is generally the case, and most module do this. For the specific case of the switch interface, I guess it is something that might happen a lot, that is why I asked you about the watchdog !
Were we saying the same thing or is there two opposite opinions here ? I'm not sure !

@Neverhorst 

Thank you for these very interesting explanations. These are exactly the kind of things that would be nice in an ADR (#586)

I was indeed only talking about hardware/logic link, as it is what interfaces are about.

1. I was not aware hardware and logic could communicate via Qt signal, I have not seen an interface doing this. Would such design be compatible with rpyc ? (Should every hardware-logic link be compatible with rpyc ?)

2. I like the solution where the interface explicitly says an error should be raised if necessary, which would of course be caught by logic and reported via self.log.error() if needed. I know this is not the current design, but I don't see where it would mess up the execution flow. I personally find it more pythonic, but I guess this is not an exact science !

3. About the setter value, I advocate for the logic to explicitly query the value after setting it (if logic deems it necessary). I think it simpler and quite elegant.  It make the code more readable and let the logic decide what is best in the situation.

This design discussion might be a bit out of scope for this PR, but it helps me to hear the opinions of people with more experience on these architecture decisions, and maybe it can generate some new ideas =)

kay-jahnke:
@alrik-durand 
I think we were not saying the same thing. Even if the hardware does not keep the state explicitly and the logic does, there is still a mismatch between the state the logic is showing and the real hardware state.
As far as I see it, in qudi the logic might keep some internal process parameters, but any measurements (and I would count the determining of the switch state as a measurement) should actually query the hardware. 

My point is not that the hardware is not allowed to be buffered nut rather that if the hardware needs this kind of buffering, it should also be done in the hardware module or in the interfuse. Only this ensures that I would still be able to query my switch state (very) fast for another hardware module that allows for fast querying.

I have added an interfuse, that achieves exactly what you want.

kay-jahnke:
@alrik-durand and @Neverhorst 
I'm happy with the current state of this PR now. Could you please give it another review.

meiwud334:
Hi all, 

An error came to me while loading the hardware module digital_switch_ni:

names_of_states must be a list of length 2, with the elements being a list of two or more names for the states.
Error during activation
  File "C:\software\qudi_n\core\module.py", line 98, in wrap_event
    base_event(*args, **kwargs)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 311, in fn
    self.transition()
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 306, in _tran
    self._after_event(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 338, in _after_event
    return getattr(self, fnname)(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 95, in _callback
    return func(obj, *args, **kwargs)
  File "C:\software\qudi_n\core\module.py", line 223, in __load_status_vars_activate
    self.on_activate()
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 126, in on_activate
    self.states = {name: self._names_of_states[name][0] for name in self._names_of_switches}
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 126, in <dictcomp>
    self.states = {name: self._names_of_states[name][0] for name in self._names_of_switches}
TypeError: list indices must be integers or slices, not str

kay-jahnke:
@meiwud334 Thanks for the tests.

The error was in the default config and the first line of the error message was the relevant one.
Please pull, modify your config and try again.

meiwud334:
Hi @kay-jahnke Thanks for the reply!

I tried with the new config in your last commit, but the loading was still unsuccessful. Here is the error messages:


Error during activation
  File "C:\software\qudi_n\core\module.py", line 98, in wrap_event
    base_event(*args, **kwargs)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 311, in fn
    self.transition()
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 306, in _tran
    self._after_event(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 338, in _after_event
    return getattr(self, fnname)(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 95, in _callback
    return func(obj, *args, **kwargs)
  File "C:\software\qudi_n\core\module.py", line 223, in __load_status_vars_activate
    self.on_activate()
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 128, in on_activate
    self.states = {name: self._names_of_states[name][0] for name in self._names_of_switches}
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 206, in states
    self._states[switch] = state
TypeError: 'NoneType' object does not support item assignment

kay-jahnke:
@meiwud334 
that was a nasty one. Please check again.

meiwud334:
@kay-jahnke Still error while loading, it seems the card doesn't like ":".

Error during activation
  File "C:\software\qudi_n\core\module.py", line 98, in wrap_event
    base_event(*args, **kwargs)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 311, in fn
    self.transition()
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 306, in _tran
    self._after_event(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 338, in _after_event
    return getattr(self, fnname)(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 95, in _callback
    return func(obj, *args, **kwargs)
  File "C:\software\qudi_n\core\module.py", line 223, in __load_status_vars_activate
    self.on_activate()
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 129, in on_activate
    self.states = {name: self._names_of_states[name][0] for name in self._names_of_switches}
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 213, in states
    with nidaqmx.Task('NISwitchTask' + self.name) as switch_task:
  File "C:\Anaconda3\envs\qudi\lib\site-packages\nidaqmx\task.py", line 96, in __init__
    check_for_error(error_code)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\nidaqmx\errors.py", line 127, in check_for_error
    raise DaqError(error_buffer.value.decode("utf-8"), error_code)
nidaqmx.errors.DaqError: Specified string is not valid, because it contains an invalid character.
Position of Invalid Character: 36
Invalid String: NISwitchTaskNICard Dev1 port0 line30:31

Task Name: NISwitchTaskNICard Dev1 port0 line30:31

Status Code: -200552
Exception ignored in: <object repr() failed>
  File "C:\Anaconda3\envs\qudi\lib\site-packages\nidaqmx\task.py", line 105, in __del__
    'reserved.'.format(self.name), DaqResourceWarning)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\nidaqmx\task.py", line 156, in name
    check_for_error(size_or_code)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\nidaqmx\errors.py", line 127, in check_for_error
    raise DaqError(error_buffer.value.decode("utf-8"), error_code)
nidaqmx.errors.DaqError: Task specified is invalid or does not exist.
Status Code: -200088

kay-jahnke:
@meiwud334 interesting, please try again.

meiwud334:
@kay-jahnke Now the channels can be read by the task yay :)! Just need to adjust the input data now.

Remote connection not secured! Use a certificate!
Error during activation
  File "C:\software\qudi_n\core\module.py", line 98, in wrap_event
    base_event(*args, **kwargs)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 311, in fn
    self.transition()
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 306, in _tran
    self._after_event(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 338, in _after_event
    return getattr(self, fnname)(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 95, in _callback
    return func(obj, *args, **kwargs)
  File "C:\software\qudi_n\core\module.py", line 223, in __load_status_vars_activate
    self.on_activate()
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 133, in on_activate
    self.states = self._states
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 218, in states
    switch_task.write(binary, auto_start=True)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\nidaqmx\task.py", line 1222, in write
    number_of_channels, 1)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\nidaqmx\task.py", line 1101, in _raise_invalid_write_num_chans_error
    DAQmxErrors.WRITE_NUM_CHANS_MISMATCH.value, task_name=self.name)
nidaqmx.errors.DaqError: Write cannot be performed, because the number of channels in the data does not match the number of channels in the task.

When writing, supply data for all channels in the task. Alternatively, modify the task to contain the same number of channels as the data written.

Number of Channels in Task: 2
Number of Channels in Data: 1

Task Name: NISwitchTaskNICard Dev1 port0 line30 31

kay-jahnke:
@meiwud334 and again please.

meiwud334:
@kay-jahnke Getting close we are.

Error during activation
  File "C:\software\qudi_n\core\module.py", line 98, in wrap_event
    base_event(*args, **kwargs)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 311, in fn
    self.transition()
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 306, in _tran
    self._after_event(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 338, in _after_event
    return getattr(self, fnname)(e)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 95, in _callback
    return func(obj, *args, **kwargs)
  File "C:\software\qudi_n\core\module.py", line 223, in __load_status_vars_activate
    self.on_activate()
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 133, in on_activate
    self.states = self._states
  File "C:\software\qudi_n\hardware\switches\digital_switch_ni.py", line 219, in states
    switch_task.write(binary, auto_start=True)
  File "C:\Anaconda3\envs\qudi\lib\site-packages\nidaqmx\task.py", line 1247, in write
    DAQmxErrors.UNKNOWN.value, task_name=self.name)
nidaqmx.errors.DaqError: Write failed, because this write method only accepts boolean samples when there is one digital line per channel in a task.

Requested sample type: <class 'int'>

Task Name: NISwitchTaskNICard Dev1 port0 line30 31

kay-jahnke:
@meiwud334 OK, made the cast explicit now. Please try again.

alrik-durand:
I have tested the OSW12 current module and there are a few problem :
- The names_of_states config generates an error, like digital_switch.
- The following code is probably unnecessary and cause bug when settin index one : 
`        if direction:
            self.log.error('No state to set.')
            return
`
- The third problem is quite mysterious : 
If I solve the problem mentioned above, the logic seems to work but the GUI does not appear. A new window seems to have been created (from Window's program bar) but it's all white (when the mouse hover the qudi icon) and I can not open it normally. I have no error in the logger so I don 't know what to do to solve it.


As a general comment, I think the first problem is the perfect example of what is wrong with inserting general features in each hardware file : this leads to copy pasting, which is bad and propagates errors. The use of user friendly names should be handled once by logic and not by each hardware, which are fine with index to me. I don't think this is C looking, this is just simple.

I also tried using the current logic for scripting and it's quite painful : I can not simply change the state of one switch. This is typical of the tendency to put every feature in the hardware and having a logic that does no check, adds no feature and simply calls directly hardware. Even if I like the new GUI, and the old code was quite bad, from a coding point of view, I feel this PR is a step in the wrong direction. Again I have no authority, but please take this opinion under consideration.

kay-jahnke:
@alrik-durand Thanks for testing.

- The first error should be fixed.
- The second one is actually very useful. As addressing states via indices is not supported anymore it tells you that you have incorrectly addressed the state and therefore no change to the state was performed.
- This error is quite weird. The only thing I can think of is a lingering app_status. Could you please delete your app_status for everything that has to do with the switches and try again.

Concerning the general points: 
- The discussion lead to everything now being addressed by dicts which have string keys, so indices are not supported anymore. Everything else would have made the hardware files quite complicated. What I could do, would be to reintroduce the indices at logic level, to make it easier to address there. Would that help you?
- A have to say I don't understand your last point. Why would it be better if the logic throws the error instead of the hardware file?

alrik-durand:
1. Thanks

2. Direction is computed at line 191 as the index of the list element. If it's 1 (50% of the time) it generates an error when casted to boolean. Please check again !

3. Indeed you were right, it solved it.

I'm not against the use of key totally, I'm against every hardware module having to copy paste the code handling this. Maybe the hardware can use a hardware specific (but generally fixed) string that is converted into a user customizable string by the logic ?

This is an issue I've noticed a lot and that too many people ignore. Logic should handle the code that is general to every hardware so that every hardware is not required to do so. This simplify code and prevent copy pasting and code repetition. This is not just laziness, this is good software practice. User scripts (where most errors generally are) should always talk to logic, so doing the checking in the logic generally enough. (Only fragile hardware module should double check to be sure.)
My previous point is directly related to this as well : as the checks are repeated, so are the features letting the config file customize the labels.

I'm just worried that someday qudi will have 100 hardware module for switches which will all recode the same features.

kay-jahnke:
@alrik-durand oops, number 2 was my fault

meiwud334:
@kay-jahnke I tested today on a radiant dye flip mirror driver with NI PCIe-6323. Every thing (gui, logic and hardware) loads fine. One thing that was noticed, that might worth mentioning is that the flip mirror trigger is only responsive to a rising edge/ low level to high level switch. A typical scenario: 

switch to high level: mirror flips up, 
switch to low level: nothing happens, 
switch to high level again: mirror folds.
switch to low level again: nothing happens.

It might be a good idea to put a bit of though into the design of logic/gui so that the state of the mirror can be also addressed together with the trigger?

kay-jahnke:
@meiwud334 
If you want to use the NI Card to trigger an external switch, please use the trigger interface (see PR #629 ) and not the switch interface.
The radiant dye flipmirror should also have a RS232 connection, which can be driven by the switch interface implemented in flipmirror.py

kay-jahnke:
@Neverhorst thanks for adjusting the interface as an outcome of all the discussions above.

@alrik-durand This should now be much more to your liking. We also discussed and concluded that as the hardware is not called by signals, we can also raise errors and than make the logic handle them accordingly.

@meiwud334 @ksenkalla @GRexBayer @jochen-scheuer @alrik-durand Please go ahead and test again.

Neverhorst:
My GUI tweaks were still missing but are now merged into this branch. Testing can commence...

alrik-durand:
@kay-jahnke Thank you for the consideration, indeed I think this is an improvement. 
I would go even further into module organization but maybe an example of what I'm proposing is better than long explanation. I've pushed on my repo a WIP I've done starting from your branch. It is not finished yet (I have to pause for today) but it gives an idea.
In a few word I'm proposing that the logic translate the names used by the hardware (which are strings specific to the hardware, but not configurable) to configurable names used by the logic and gui specified by config.
To the GUI or scripts there is no difference, but for hardware modules things are much simpler.

I'll try to finish what I've started in the next few days.

Neverhorst:
@alrik-durand Well, I really do not share the opinion that the names should be defined by the logic. In Python it makes no difference (in most cases) if you address something by name or by index. We just chose "by name" because it makes it more verbose and easy to understand/read (which is a pillar of qudi and Python in general). Also you do not need to keep both the name and index information (lesser complexity).
Additionally all modules using a specific hardware module share the same information by default if the hardware is defining the names, which is the more direct and less confusing approach compared to multiple modules defining different names for the same hardware switch.
The argument about code duplication in activation is now invalidated since it is all wrapped in an interface method.
Also nobody is forcing you to make the names a `ConfigOption` in the hardware modules; you could just make them hard coded for simple hardware.
All in all the hardware modules are really lean now and just define the basics.

kay-jahnke:
@Neverhorst 
it seems the yaml representer has some problems with the new Enums you defined in the GUI:

```
Failed to save status variables of module gui.switches:
OrderedDict([('pos_x', 632), ('pos_y', 249), ('switch_style', <SwitchStyle.TOGGLE_SWITCH: 0>), ('colorscheme', <ColorScheme.DEFAULT: 0>), ('alt_toggle_switch_style', False)])
  File "C:\Software\qudi\core\manager.py", line 1292, in saveStatusVariables
    config.save(filename, variables)
  File "C:\Software\qudi\core\config.py", line 250, in save
    ordered_dump(data, stream=f, Dumper=yaml.SafeDumper, default_flow_style=False)
  File "C:\Software\qudi\core\config.py", line 227, in ordered_dump
    return yaml.dump(data, stream, OrderedDumper, **kwds)
  File "C:\Software\Anaconda3\envs\qudi\lib\site-packages\ruamel\yaml\main.py", line 849, in dump
    version=version, tags=tags, block_seq_indent=block_seq_indent)
  File "C:\Software\Anaconda3\envs\qudi\lib\site-packages\ruamel\yaml\main.py", line 810, in dump_all
    dumper._representer.represent(data)
  File "C:\Software\Anaconda3\envs\qudi\lib\site-packages\ruamel\yaml\representer.py", line 73, in represent
    node = self.represent_data(data)
  File "C:\Software\Anaconda3\envs\qudi\lib\site-packages\ruamel\yaml\representer.py", line 101, in represent_data
    node = self.yaml_representers[data_types[0]](self, data)
  File "C:\Software\qudi\core\config.py", line 165, in represent_ordereddict
    dict_data.items())
  File "C:\Software\Anaconda3\envs\qudi\lib\site-packages\ruamel\yaml\representer.py", line 206, in represent_mapping
    node_value = self.represent_data(item_value)
  File "C:\Software\Anaconda3\envs\qudi\lib\site-packages\ruamel\yaml\representer.py", line 111, in represent_data
    node = self.yaml_representers[None](self, data)
  File "C:\Software\Anaconda3\envs\qudi\lib\site-packages\ruamel\yaml\representer.py", line 375, in represent_undefined
    raise RepresenterError("cannot represent an object: %s" % data)
ruamel.yaml.representer.RepresenterError: cannot represent an object: SwitchStyle.TOGGLE_SWITCH
```

Neverhorst:
> @Neverhorst
> it seems the yaml representer has some problems with the new Enums you defined in the GUI:

Ok, that's weird. `IntEnum` is a subclass of `int` which is why I expected the representer to identify it as such. Seems like the representer is checking type using `type()`. This is why you should (almost) always use `isinstance()`.
I will look into it.

Neverhorst:
@kay-jahnke 
Introduced a quick workaround for yaml. Also fixed a bug with saving and restoring switch states.

kay-jahnke:
@Neverhorst thanks for the fix. I tested and it works for me now.

alrik-durand:
@Neverhorst As I've proposed on my demo, the hardware should use names as indeed it is more explicit than index.
Any real hardware has a convention anyway on what the states are : 1/2, High/Low, A/B, etc. It seems natural to let the hardware module keep this convention as it help understanding the mapping. If we want to have a more explicit naming, the user is creating a second convention and that's ok - we just need to define the mapping by specifying the config. I think this is complex not complicated.
I agree code duplication can be partially resolved by using a trick with the interface, but this violate the division into hardware, logic, and interface a lot. If some logic checks is to be done to every hardware module, it should be done by logic.

With that said, I let you decide what is best.

GRexBayer:
I tried running the SwitchGUI from the "qudi: Manager" on my office Computer (Win 10) with the default config. 
It shows me the following error message:

Error during activation
AssertionError: Invalid switch state: "eins"

  File "C:\Program Files\Anaconda3\Lib\site-packages\qudi\core\module.py", line 98, in wrap_event
    base_event(*args, **kwargs)

  File "C:\Program Files\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 311, in fn
    self.transition()

  File "C:\Program Files\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 306, in _tran
    self._after_event(e)

  File "C:\Program Files\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 338, in _after_event
    return getattr(self, fnname)(e)

  File "C:\Program Files\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 95, in _callback
    return func(obj, *args, **kwargs)

  File "C:\Program Files\Anaconda3\Lib\site-packages\qudi\core\module.py", line 223, in __load_status_vars_activate
    self.on_activate()

  File "C:\Program Files\Anaconda3\Lib\site-packages\qudi\gui\switch\switch_gui.py", line 149, in on_activate
    self._switches_updated(self.switchlogic().states)

  File "C:\Program Files\Anaconda3\Lib\site-packages\qudi\gui\switch\switch_gui.py", line 237, in _switches_updated
    self._widgets[switch][1].set_state(state)

  File "C:\Program Files\Anaconda3\Lib\site-packages\qudi\gui\switch\switch_state_widgets.py", line 73, in set_state
    assert state in self.switch_states, f'Invalid switch state: "{state}"'

Perhaps I messed up somewhere?

GRexBayer:
I think this is just a minor detail, but upon reopening, the module remembers all the view settings for the Toggle bars (whether one had toggle bars or not).
But it does not remember the color scheme (red-green)

Neverhorst:
> I think this is just a minor detail, but upon reopening, the module remembers all the view settings for the Toggle bars (whether one had toggle bars or not).
> But it does not remember the color scheme (red-green)

This should be fixed now.

GRexBayer:
Checked the GUI again on Windows 10 OS with dummy hardware.
Under the possible operations I could think of, the program runs smooth. Unless I clean the Module status file, it remembers the preferences for the button style. Reloading the config file with the module deactivated or activated seems to result in no error.

Yawghmoth:
Tested flip-mirror (radiant dyes box) which is now working after small changes on the flip-mirror hardware file

Yawghmoth (approved):
Tested with an Radiant dyes flippmirror, which worked so far

Yawghmoth (commented):
changed ask to querry (not add)

jochen-scheuer (approved):
Tested it on two different setups and works as intended.

PR #630, opened 2020-11-17
Predefined methods bugfix

Geegee:
The following generate methods have changed:

Ramsey, hahnecho, hahnecho_exp, xy8_tau, xy8_freq.
## Description

The changes fall into two categories:
    i) In the case that some or all tau values would lead to
       overlapping pulses the minimal tau is adjusted.
       For hanecho_exp as well as xy8_freq, the minimum tau is
       adjusted and the step size, for all others only the
       minimum tau is adjusted leading to a larger final tau.
    2) In the XY8 methods the tau half free evolution elements
       do not consider the different time between the center  of consecutive pulses i.e. (pi / 2, pi) vs. (pi, pi)
       vs. (pi, 3pi / 2)
## Motivation and Context

Choosing wrong free evolution for the $\frac{\tau}{2}$ elements in xy8 type measurements.
Generate methods behave more predictable (no points are recorded twice without notice)
Avoiding measuring points twice. 

## How Has This Been Tested?

Generating waveforms for all the methods with analog, not analog, alternating and non alternating 
settings. Checking for each the interpulse spacing and pulse lengths. 

## Types of changes

## Checklist:

### DIFF
--- a/logic/pulsed/predefined_generate_methods/basic_predefined_methods.py
+++ b/logic/pulsed/predefined_generate_methods/basic_predefined_methods.py
@@ -272,7 +272,8 @@ class BasicPredefinedGenerator(PredefinedGeneratorBase):
         created_sequences = list()
 
         # get tau array for measurement ticks
-        tau_array = tau_start + np.arange(num_of_points) * tau_step
+        real_start_tau = max(0, tau_start - self.rabi_period / 4)
+        tau_array = real_start_tau + np.arange(num_of_points) * tau_step
 
         # create the elements
         waiting_element = self._get_idle_element(length=self.wait_time,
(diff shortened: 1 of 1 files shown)


timoML:
For the general audience, here's what @Geegee and me discussed:
- put the logic for converting and checking taus into a seperate helper method 
- this PR would change the definition of taus in HE and Ramsey from "pulse spacing between MW" to "time between centers of MW pulses" (just like it's already defined in generate_xy8()). While we think this is more reasonable, we should discuss how to release and document this change.

Neverhorst:
@timoML @Geegee Is this PR still valid after our discussions regarding the definition of inter-pulse spacings?

timoML:
This PR will be replaced by some upcoming one based on the advancedsequences branch. Can be closed from my side.

timoML:
This should have become obsolte with the warnings introduced in #650. @Geegee: Can you confirm, then we can close?

Geegee:
Yes can be closed now.

PR #636, opened 2020-12-14
Cryomagnetics vector magnet power supply hardware

alrik-durand:
## Description
This PR adds a hardware module to interface Cryomagnetics vector magnet power supply. No interface is used but hardware can be operated via notebooks.

## Motivation and Context
This PR adds support for Cryomagnetics vector magnet power supply. 
It does not use the MagnetInterface as it is an old interface very hardware specific which is not well suited for this hardware. Developing a proper interface, logic and GUI for a superconducting magnet should be done at a later stage.
This PR is enough to interact via script for the moment.

## How Has This Been Tested?
This has been tested with the hardware via ethernet via a notebook.

## Types of changes

## Checklist:


alrik-durand:
I just updated the code with a new version.
This update was required because setting a setpoint in Gauss create problem because the hardware is rounding the value to the nearest gauss integer (why I don't know...).
This update uses the current as setpoint to circumvent the problem.

PR #637, opened 2020-12-16
Adding the possibility to disable the scrolling of the mouse wheel in…

jochen-scheuer:
Adding the possibility to disable the scrolling of the mouse wheel in scientific spinboxes.

## Description

Adding the possibility to disable the scrolling of the mouse wheel in scientific spinboxes. This sometimes needed when scrolling through a gui is desired and valus should not be changed by accident.

## How Has This Been Tested?

Tested in dummy. 

## Types of changes

## Checklist:


kay-jahnke (approved):
tested on dummy with Win10 and my own modules: it works well

Neverhorst (approved):
Works, but keep in mind that this will not prevent the SpinBox from gaining focus via mouse wheel. It will only prevent value changes via wheel.

Issue #633, opened 2020-12-01
implement pulse or sequence file size alert. 


liuyanulm:
## *What* is affected by this bug?

If the pulse file is too large (comparing to the the capability of AWG), AWG would crash, or the connection between AWG and lab-pc would crash. This case happens not rarely with Tek 70k.

## *When* does this occur?

If the pulse file is too large (comparing to the the capability of AWG), remind user to review the pulse settings.

timoML:
By now, the sequence generation should yield an error and stop if the maximum number of samples (defined in get_constraints().waveform_length.max) is exceeded.

Issue #551, opened 2019-09-20
Please disable the "Run pulsedmeasurement Button", during pulser generation


liuyanulm:
## *What* is affected by this bug?

During the pulser GenLoadSamp process, I sometimes mis-click on the "Run pulsed-measurement" and this leads to a serious problem which would break the connection with AWG-70k. And that is not the finish, I have to reset everything on the AWG, the ip, the ftp, etc., to rebuild connection with Qudi.
Leave a comment if you have the same problem, or not.

## *When* does this occur?

mis-click on the "Run pulsed-measurement", during the pulser GenLoadSamp process.

## *Where* on the platform does it happen?

Windows 10

## Other Comments

I guess we only need to disable the "Run" button as how the "Pulser On", "GenLoaSam" were disabled..

PR #639, opened 2020-12-17
Cosmetic changes to waveform upload

timoML:
This introduces improvements to the upload process of waveforms and the pulsed gui.

## Description
1. Better visual feedback on an ongoing sampling / loading process
2. Disable the start measurement button while busy
3. Estimate the expected upload time
4. Abort and log error if waveform length exceeds the max defined in device constraints.

As we discussed (3) already, I assume that the solution introduced here is rather controversial.
I put this (quick and simple) possible solution up for continuing the discussion here:
- On the logic level, the sequencegenerator keeps track of the experienced durations it takes to finish pulsegenerator().write_waveform() and takes them as benchmarks (these durations are logged right now anyway).
- It will then estimate the time a number of samples will need for upload. If this time exceeds a configured value, it will issue an info including the estimated finish time.
- Of course the acquired benchmarks are only valid for a specific pulse generator device. Thus, the sequencegenerator creates a hash from the used device (or rather: its constraints) plus its configuration and resets the benchmarks when this hash is altered. 
I guess there could be scenarios where the upload speed of a device changes while the hash stays constant. 
However, this is rather unlikely and will only affect the accuracy of an info to the user.

## Motivation and Context
In response to issues #633, #551 

## How Has This Been Tested?
1. & 2: on setup 3 for some hours
3. I uploaded waveforms to my Keysight M8190a and checked the estimated write time with the real one. The accuracy seems to be within ~ 20 %.  

## Screenshots:
image
image

## Types of changes:

## Checklist:


logic/pulsed/sequence_generator_logic.py
```diff
@@ -1623,6 +1632,60 @@ def _sampling_sequence_sanity_check(self, sequence):
...
+        Creates a unique id for the currently used pulse generator.
+        :return: pg_id
+        """
+
+        # indirect creation, as the pulser interface doesn't provide eg. the serial number or model of the device
+        # based on 1. ScalarConstraints in device constraints, 2. config options of device
+        constraint_dict = OrderedDict(
+            (key, val.__dict__) for (key, val) in self.pulsegenerator().get_constraints().__dict__.items() if
+            isinstance(val, ScalarConstraint))
+        id_dict = self.pulsegenerator()._configuration.update(constraint_dict)
```
timoML, line 1650: Accessing _configuration might be a bit ugly, but the BaseMixin guarantees that it exists.

Neverhorst: accessing a private variable from outside the module is always a big no-no and the attribute `_configuration` will not exist in the new qudi core.
Apart from that this line will always cause `id_dict == None`.

timoML: I agree with the private variable. If we agree that this quick solution is something we want to follow on, we could provide read access to the config or implement unique device / config ids.
Don't see though, why the dict shouldn't work? It does here...

timoML: Fixed the latter.
I agree on the private variable. If we want to continue on the upload time estimation in a similar way, we could either provide read access or implement unique device/config ids.

gui/pulsed/pulsed_maingui.py
```diff
@@ -389,6 +390,11 @@ def _connect_logic_signals(self):
         self.pulsedmasterlogic().sigMeasurementSettingsUpdated.connect(self.measurement_settings_updated)
         self.pulsedmasterlogic().sigAnalysisSettingsUpdated.connect(self.analysis_settings_updated)
         self.pulsedmasterlogic().sigExtractionSettingsUpdated.connect(self.extraction_settings_updated)
+        self.pulsedmasterlogic().sigSampleBlockEnsemble.connect(self.sampling_or_loading_busy)
```
timoML, line 393: Something is wrong in disconnection. Causes error when unloading and reloading pulsed gui.

timoML: Ok, seems to be unrelated to this PR but an issue in the current master (tested on 2 setups):
In the manager hit the buttons:
- "deactivate module"
- "Load this module and all dependencies" (NOT the reload button)

RuntimeError: wrapped C/C++ object of type QComboBox has been deleted
  File "C:\Users\Setup3-PC\Desktop\qudi\core\module.py", line 98, in wrap_event
    base_event(*args, **kwargs)
  File "C:\ProgramData\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 311, in fn
    self.transition()
  File "C:\ProgramData\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 306, in _tran
    self._after_event(e)
  File "C:\ProgramData\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 338, in _after_event
    return getattr(self, fnname)(e)
  File "C:\ProgramData\Anaconda3\envs\qudi\lib\site-packages\fysom\__init__.py", line 95, in _callback
    return func(obj, *args, **kwargs)
  File "C:\Users\Setup3-PC\Desktop\qudi\core\module.py", line 223, in __load_status_vars_activate
    self.on_activate()
  File "C:\Users\Setup3-PC\Desktop\qudi\gui\pulsed\pulsed_maingui.py", line 193, in on_activate
    self._activate_generator_settings_ui()
  File "C:\Users\Setup3-PC\Desktop\qudi\gui\pulsed\pulsed_maingui.py", line 944, in _activate_generator_settings_ui
    self.pulse_generator_settings_updated(self.pulsedmasterlogic().pulse_generator_settings)
  File "C:\Users\Setup3-PC\Desktop\qudi\gui\pulsed\pulsed_maingui.py", line 1378, in pulse_generator_settings_updated
    widget.blockSignals(True)

timoML:
After discussion with @Neverhorst, re-wrote the benchmarking part.
Now benchmarks
- predict upload times way more exact (Delta ~5 % for my setup)
- can be triggered by the user in gui
- or come from a rolling buffer of experienced uploads
- don't detect hardware/config changes anymore
- are run on both writing and loading
- if none available, will issue a warning on pulsed gui startup

logic/pulsed/sequence_generator_logic.py
```diff
@@ -2071,3 +2238,169 @@ def _delete_sequence(self, names):
...
+        def round_to_granularity(n_samples):
+            granularity = self.pulsegenerator().get_constraints().waveform_length.step
+            return np.ceil(n_samples / granularity) * granularity
+
+        self._benchmark_write.reset()
+        self._benchmark_load.reset()
+
+        waveform_name = 'qudi_benchmark_chunk'
+        n_init_guess = [1e4, 1e6]
+        n_max_fix = 10e6
```
timoML, line 2254: These hard-coded parameters limit the generality. Open for suggestions, but hard to guess reasonable benchmark chunk sizes without knowing the coarse speed of the device.

gui/pulsed/pulsed_maingui.py
```diff
@@ -206,6 +207,36 @@ def on_activate(self):
         self._connect_logic_signals()
 
         self.show()
+
+        if np.isnan(self.pulsedmasterlogic().sequencegeneratorlogic().get_speed_write_load()):
+            dialog = QtWidgets.QDialog(self._mw)
```
Neverhorst, line 212: You are creating a `QWidget` object here and giving it a parent `QObject` (`self._mw`).
If you are not explicitly destroying this dialog object after use, a reference will remain in the child tree of `self._mw` and it will not get garbage collected.
I see 2 ways to solve this:

1) explicitly set the parent to `None` at the end of this method

yet better

2) Use [`QMessageBox`](https://doc.qt.io/qt-5/qmessagebox.html#details) instead (also saves a lot of unnecessary code that is becoming obsolete in the new qudi core)

gui/pulsed/pulsed_maingui.py
```diff
@@ -3103,4 +3172,8 @@ def update_laser_data(self):
         self.lasertrace_image.setData(x=x_data, y=y_data)
         return
 
+    @QtCore.Slot()
+    def run_pg_benchmark(self):
 
+        self.pulsedmasterlogic().sequencegeneratorlogic().sigRunPgBenchmark.emit()
+        self.pulsedmasterlogic().sigGeneratorSettingsChanged.emit({})
```
Neverhorst, line 3179: always add newline at end of file

gui/pulsed/pulsed_maingui.py
```diff
@@ -3103,4 +3172,8 @@ def update_laser_data(self):
         self.lasertrace_image.setData(x=x_data, y=y_data)
         return
 
+    @QtCore.Slot()
+    def run_pg_benchmark(self):
 
+        self.pulsedmasterlogic().sequencegeneratorlogic().sigRunPgBenchmark.emit()
```
Neverhorst, line 3178: This is an absolute no-no. Never emit signals from outside the QObject they belong to (see also comment in logic)

Neverhorst (changes requested):
The whole thing blows up in my face when I try to run that on dummy config:
```
2021-01-26 17:03:00 error hardware.pulser_dummy.PulserDummy Loading failed. Analog channel 3 not active.
2021-01-26 17:03:00 error root 
  File "C:\Software\qudi\logic\pulsed\sequence_generator_logic.py", line 2265, in run_pg_benchmark
    persistent_datapoint=True)
  File "C:\Software\qudi\logic\pulsed\sequence_generator_logic.py", line 2382, in _sample_load_benchmark_chunk
    loaded_dict = self.pulsegenerator().load_waveform(wfm_list)[0]
KeyError: 0
```

This is actually worse than it seems. 
Since this is obviously not happening at your setup, this suggests you are running a different version of qudi and/or have fiddled with the pulser interface since this error happens due to a non-interface-conform return value.

Anyway if I correct this error the benchmark continues but fails again when trying to load the waveform calling `load_waveform` in the dummy hardware. 

Code cleaning is needed as well as much more testing of the benchmark routine in dummy and on other hardware.

gui/pulsed/pulsed_maingui.py
Neverhorst, line 3179: Why is this signal emitted here?

Neverhorst (changes requested):

gui/pulsed/pulsed_maingui.py
timoML: Thanks for the in-depth review. I will have a look at the issues raised.
I agree, the benchmark tool should probably live outside the sequence generator logic, but I didn't find an obvious well-suited place. Any suggestion?

logic/pulsed/sequence_generator_logic.py
```diff
@@ -2071,3 +2238,172 @@ def _delete_sequence(self, names):
...
+
+        if written_samples != n_samples:
+            self.log.error('Sampling of benchmark chunk failed. '
+                           'Write to device was unsuccessful.\nThe number of '
+                           'actually written samples ({:d}) does not match '
+                           'the number of samples staged to write ({:d}).'
+                           ''.format(written_samples,
+                                     n_samples))
+
+        if not self.__sequence_generation_in_progress:
```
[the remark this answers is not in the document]
timoML: See your point and removed the check for __sequence_generation_in_progress. I just copied the behavior in sample_pulse_block_ensemble(). Should we unlock unconditionally there, too?

logic/pulsed/sequence_generator_logic.py
```diff
@@ -41,6 +43,119 @@
 from logic.pulsed.sampling_functions import SamplingFunctions
 from interface.pulser_interface import SequenceOption
 
+from collections import deque, namedtuple
+import scipy
+
+
+Datapoint = namedtuple('Datapoint', 'time_s y')
```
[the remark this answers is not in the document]
timoML: It helped me during debugging, as I had mixed up the axes.^^ But you're right, it's probably not worth the serializing/de-serializing effort.

gui/pulsed/pulsed_maingui.py
timoML: Switched to QMessageBox. Minor disadvantage: as far as I see, there is no way to center the button position. This seems to be style for other qudi dialogues. Now it looks like:
image

logic/pulsed/sequence_generator_logic.py
```diff
@@ -199,6 +324,10 @@ def on_activate(self):
         self._pog = PulseObjectGenerator(sequencegeneratorlogic=self)
 
         self.__sequence_generation_in_progress = False
+
+        self.sigRunPgBenchmark.connect(
```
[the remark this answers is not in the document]
timoML: I think I fixed this by defining own signals in pulsed_maingui.py.

timoML:
Fixed the blow-up on dummy hardware (caused by not setting the activation config correctly) and all other issues mentioned. Could you give it a second look?

Neverhorst (changes requested):
I get an error when using this change on my setup with Tektronix AWG70002A (2 channels):
```
  File "C:\Software\qudi_private\logic\pulsed\sequence_generator_logic.py", line 2148, in run_pg_benchmark
    persistent_datapoint=True)
  File "C:\Software\qudi_private\logic\pulsed\sequence_generator_logic.py", line 2235, in _sample_load_benchmark_chunk
    pg_chs_a, pg_chs_d = _get_largest_channel_config()
  File "C:\Software\qudi_private\logic\pulsed\sequence_generator_logic.py", line 2219, in _get_largest_channel_config
    a_chs = sorted([ch for ch in configs[largest_config] if ch.startswith('a_')])
KeyError: frozenset({'a_ch1', 'd_ch2', 'd_ch3', 'd_ch4', 'd_ch1', 'a_ch2'})
```
This happens due to a bug I just fixed when there is the activation config "all" present in the hardware constraints. 

After I had fixed this bug, I get the following warning messages during the initial benchmark:
```
Loading of waves ['qudi_benchmark_chunk_ch1', 'qudi_benchmark_chunk_ch2'] failed, still: {1: 'qudi_benchmark_chunk_ch1', 2: 'qudi_benchmark_chunk_ch1'}
Loading of waves ['qudi_benchmark_chunk_ch1', 'qudi_benchmark_chunk_ch2'] failed, still: {1: 'qudi_benchmark_chunk_ch1', 2: 'qudi_benchmark_chunk_ch1'}
Loading of waves ['qudi_benchmark_chunk_ch1', 'qudi_benchmark_chunk_ch2'] failed, still: {1: 'qudi_benchmark_chunk_ch1', 2: 'qudi_benchmark_chunk_ch1'}
Loading of waves ['qudi_benchmark_chunk_ch1', 'qudi_benchmark_chunk_ch2'] failed, still: {1: 'qudi_benchmark_chunk_ch1', 2: 'qudi_benchmark_chunk_ch1'}
Loading of waves ['qudi_benchmark_chunk_ch1', 'qudi_benchmark_chunk_ch2'] failed, still: {1: 'qudi_benchmark_chunk_ch1', 2: 'qudi_benchmark_chunk_ch1'}
Loading of waves ['qudi_benchmark_chunk_ch1', 'qudi_benchmark_chunk_ch2'] failed, still: {1: 'qudi_benchmark_chunk_ch1', 2: 'qudi_benchmark_chunk_ch1'}
```
Currently no idea what this actually means (a clearer message content would be nice).

In general I find it very confusing that the user is not informed of the benchmark status/progress. Maybe it would be a good thing to keep the modal dialog open and display some kind of progress report (loading bar or the moving circle thingy).
Otherwise one would be able to fiddle around in the GUI and break the benchmarking (starting/stopping pulser etc.). Also you do not know when the benchmark has been completed.

Regarding the initial guess for chunk sizes to benchmark, you could take the hardware constraints for waveform length as an indicator, e.g. start with 1% of the maximum size or something like that.

timoML:
> I get an error when using this change on my setup with Tektronix AWG70002A (2 channels):
> Loading of waves ['qudi_benchmark_chunk_ch1', 'qudi_benchmark_chunk_ch2'] failed, still: {1: 'qudi_benchmark_chunk_ch1', 2: 'qudi_benchmark_chunk_ch1'}
> ```
> 
> Currently no idea what this actually means (a clearer message content would be nice).

Well, loading of the waveform failed, because the second channel is loaded with the first channel waveform. Probably it's a misunderstaning how the pulser interfaces abstracts away multi-channel waveforms.
I'll have a look next week.

> In general I find it very confusing that the user is not informed of the benchmark status/progress. Maybe it would be a good thing to keep the modal dialog open and display some kind of progress report (loading bar or the moving circle thingy).

Currently, there is a info message about start/completion of the benchmark. You're right, a GUI element would be nicer, though.

> Regarding the initial guess for chunk sizes to benchmark, you could take the hardware constraints for waveform length as an indicator, e.g. start with 1% of the maximum size or something like that.

Like the idea.

gui/pulsed/pulsed_maingui.py
Neverhorst: You could simply move the benchmarking code to a separate file, e.g. `qudi/logic/pulsed/_pulser_benchmark.py`. It could contain a class or a function to run the benchmark that needs to be handed over the `SequenceGeneratorLogic` instance.
Something along these lines:

``` python
from ._pulser_benchmark import MyBenchmarkTool

...

class SequenceGeneratorLogic(LogicBase):
...
    # somewhere in SequenceGeneratorLogic
    def run_benchmark(self):
        tool = MyBenchmarkTool(the_logic=self)
        result = tool.run()
        ...
```

logic/pulsed/sequence_generator_logic.py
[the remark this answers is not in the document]
Neverhorst: No, in the other case the module should stay locked if sequence generation is in progress.

Neverhorst:
Seeing how this benchmarking tool is blowing up, maybe it would be a good thing to completely move everything related to a separate (non-threaded) logic module or at least a separate QObject subclass that is handling everything and just fires signals updating the status.

timoML:
I guess we're good to merge. What do you think @Neverhorst?

Notes:
- Negative speeds shouldn't happen anymore
- I intentionally pushed a commit that moved more code from the sequence generator logic into a own benchmark class. Had to revert it, because serializing the StatusVar seems hard to me if this pulse generator benchmark tool is not loaded as a module. However, as it's currently only 3 additional functions in sequence generator, I would suggest to avoid the headache and leave it as it is...

Neverhorst (approved):

PR #640, opened 2020-12-18
Ni analog output

alrik-durand:
## Description
I've added a hardware module to interface a NI9263 4 channel analog output. Hardware is interfaced via the ProcessControlInterface interface.

## Motivation and Context
Many experiment uses NI 6323 card with multiple functionalities. Analog output is one of them the card only have 4 analog output. The NI9263 can be used via USB or ethernet (depending of the module used for connection) and add more analog output.
Theses analog output are interfaced via the ProcessControlInterface.

## How Has This Been Tested?
Tested with a NI9263 connected via USB cDAQ-9171 to a windows 10 computer.

## Types of changes

## Checklist:


meiwud334 (approved):
We don't have the hardware here in Ulm, basing only on the code no problem were spotted.

QTslav (commented):
Is it necessary to set the analog output voltage to zero when deactivating the module? 
I could see some usecases where it would be preferential to keep the voltage applied.

Neverhorst:
> We don't have the hardware here in Ulm, basing only on the code no problem were spotted.

Well, of course we have the hardware. You can basically take any of the \***insert arbitrary large number**\* NI cards in our labs.

PR #641, opened 2020-12-18
Switch bugfixes

Neverhorst:
## Description
Fixed a bug in the switch interface caused by decorating not only the property but also the setter function as abstractmethod.

Removed explicit states property declaration from dummy module to showcase the working change.

Fixed a bug in the switch logic causing the watchdog to not register switch state changes.

## How Has This Been Tested?
Win10 dummy

## Types of changes

## Checklist:

### DIFF
--- a/hardware/switches/switch_dummy.py
+++ b/hardware/switches/switch_dummy.py
@@ -91,29 +91,6 @@ class SwitchDummy(Base, SwitchInterface):
         """
         return self._switches.copy()
 
-    @property
-    def states(self):
-        """ The current states the hardware is in as state dictionary with switch names as keys and
-        state names as values.
-
-        @return dict: All the current states of the switches in the form {"switch": "state"}
-        """
-        return self._states.copy()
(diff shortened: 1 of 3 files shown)


kay-jahnke (approved):
tested on dummy and saw the error. This fixes it.

One question though, what is the purpose of the wrapper `@interface_method` then?

Neverhorst:
@kay-jahnke the purpose of `@interface_method` is still that you can overload the method (or property) using qudi interfaces and `Connector` objects.
I'm not sure that this will work flawlessly with Python properties since `interface_method` was not designed to work with them (but apparently it works anyways to some extent).
I would assume it works with read-only properties but will get namespace issues if trying to define a setter method for an interface-overloaded property using the decorator approach, e.g. with `@<property_name>.setter`. Needs testing to be sure.

Issue with this particular problem was, that the property setter must not be an interface method (only the property itself which is then refering to the setter)

PR #642, opened 2021-01-15
Odmr fit on all ranges

NicolasStaudenmaier:
Adding the possibility to do fits in cw odmr not only on a single range but on the data of all ranges. 

## Description

If the Fit range is set to -1 the data of all odmr ranges is used for the fit. 

## Motivation and Context

Useful for example for fitting the NV center mS = +1 and -1 resonances in two different ranges at the same time. 

## How Has This Been Tested?

Tested with dummy on Windows 10. 

## Types of changes

## Checklist:


Geegee:
Works well, tried it both on dummy and on the widefield setup.

Geegee (approved):
Looks good.

PR #644, opened 2021-02-05
Awg70k odmr

Geegee:
Use the Tektronix AWG70k as a microwave source that works in the ODMR toolchain. 

## Description

@Neverhorst made the AWG70k comply with the microwave interface. 
I made the get_status functions use overloading. Introduced a new function that 
gets the sampling rate of the simple waveform compiler and additionally a small change
that ensures that the number of samples is even. 
## How Has This Been Tested?

Could only test with some changes as the way I run ODMR differs somewhat 
due to different hardware (mainly I need an additional sequence step that is only run once). 

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] My code follows the code style of this project.

### DIFF
--- a/hardware/awg/tektronix_awg70k.py
+++ b/hardware/awg/tektronix_awg70k.py
@@ -35,7 +35,7 @@ from core.module import Base
 from core.configoption import ConfigOption
 from core.util.modules import get_home_dir
 from core.util.helpers import natural_sort
-from interface.pulser_interface import PulserInterface, PulserConstraints
+from interface.pulser_interface import PulserInterface, PulserConstraints, SequenceOption
 from interface.microwave_interface import MicrowaveInterface, MicrowaveLimits, MicrowaveMode, TriggerEdge
 
 
@@ -346,7 +346,7 @@ class AWG70K(Base, PulserInterface, MicrowaveInterface):
             # wait until the AWG is actually running
             while not self._is_output_on():
                 time.sleep(0.25)
-        return self.get_status()[0]
+        return self.get_status['PulserInterface']()[0]
 
     def pulser_off(self):
         """ Switches the pulsing device off.
@@ -361,7 +361,7 @@ class AWG70K(Base, PulserInterface, MicrowaveInterface):
             # wait until the AWG has actually stopped
             while self._is_output_on():
                 time.sleep(0.25)
-        return self.get_status()[0]
+        return self.get_status['PulserInterface']()[0]
 
     def write_waveform(self, name, analog_samples, digital_samples, is_first_chunk, is_last_chunk,
                        total_number_of_samples):
@@ -824,6 +824,10 @@ class AWG70K(Base, PulserInterface, MicrowaveInterface):
         @return float: The current sample rate of the device (in Hz)
         """
         return_rate = float(self.query('CLOCK:SRATE?'))
+        return return_rate
+
+    def get_wave_form_compiler_sample_rate(self):
+        return_rate = float(self.query('BWAV:SRAT?'))
         return return_rate
 
     def get_analog_level(self, amplitude=None, offset=None):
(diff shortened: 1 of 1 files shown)


Neverhorst (commented):
I tested it on my setup and unfortunately it still gives me the same instabilities and extremely long waiting times as back when I tried to implement that.
First time I started an ODMR scan it took about 2 min before it actually started measuring. Then I stopped and changed the power and started again. Boom, waited for 10 min before I gave up because ODMR and the AWG did not respond anymore. Checking the socket server on the AWG itself revealed an ongoing read operation from client. This is exactly the kind of problem what makes those Tektronix AWG so bad. The server running on the AWG just deadlocks randomly on queries and the commands `*WAI` and `*OPC?` do not behave as specified in the standard.

Long story short, in my opinion this scheme of using the Tektronix AWG70k is completely unuseable. Very long waiting times during experiments, extremely volatile device communication and missing sync between this functionality and other modules using the AWG (like pulsed).

The additional cost for a cheap CW microwave device (~2k EUR) and a simple (slow) microwave switch (~100 EUR) makes a setup so much more productive and stable compared to using this integrated solution. And if you can afford a multi-100k AWG, these additional parts should be in the budget.

It's a shame that Tektronix is so extremely incompetent when it comes to their Software but it currently limits the usefulness for applications. (Analog hardware is great though)

Neverhorst:
@Geegee Do you encounter an entirely different user experience?

Geegee:
What I can say is that it is slower compared to the setup with e.g. the SMIQ, but besides that it is running fine. As far as I remember also @anjusha-vs was using it for some time. Maybe it is a problem with the 2 channel AWG? It could also
be the number of points. I usually use around 50 - 150 points, which is quiet fast (maybe 20 seconds). I just checked for 300 points it maybe takes a minute.

ksenkalla (commented):
The pull request was tested on a setup with Windows 8.1 Enterprise and with an one channel Tektronix AWG 70001A. The output was checked with an Antritsu Spectrum Analyzer.
The first loading of the ODMR module after starting qudi causes the AWG "Compiling basic waveforms" taking actually a long time. 
image

 As long as the ODMR GUI is not loaded yet loading other modules causes the whole qudi to crash. After some while the ODMR GUI appears with a timeout error notification. 
image

I am not sure if this only happens to me or was also observed in other setups. 

Another severe error is that the actual sampling rate of the hardware is not checked but assumed to be 50GS/s, even if in the config already something differnt is defined.
Loading and playing a waveform with not the proper sampling rate causes a wrong output (checked with the spectrum analyzer).

The test of the bounderies in the GUI was successful, no wrong parameter could be set. 
With configuring the sampling rate to 50GS/s the output was correct for a single frequency as well as for ODMR runs.

I tested ODMR runs up to 4k points. These runs were successful but took a lot of time to upload. For 1k points it was about 5 min, however, double the points took more than double the time.
For a common use of the ODMR, finding some unknown resonance frequency this is very unconvenient way of working.
During the upload it is not possible to change the parameters (which is basically good) or interrupt the upload. Especially considering an accidentally set input parameter, it can be useful to interrupt a maybe very long uploading process.
It is not possible to load also other modules relying on the AWG when it is still busy. 
This will cause an severe error in the communication, so that both, the AWG and qudi has to be restarted to again communicate properly.

Maybe this pull request could be an option when the bugs are fixed for only small ranges (~100 points), but for the common use of ODMR it is currently not stable enough.

Geegee:
Thanks a lot for the rigorous testing, especially the sampling rate issue. It turns out that there are a couple more
intricacies with the Tektronix AWG. First the sampling rate in the waveform compiler can not be set below a 
certain value. This value seems to be 200 / 33 (6.06...) samples per period given the selected frequency. 
The second point is that when making the waveform compiler active in the software it automatically selects 
50 GS/s as you observed. 
The second is easily fixed by setting the sampling frequency in the waveform compiler to the sampling frequency
of the AWG after making the waveform compiler plugin active. For the first one I made a check so that the highest
frequency given the sampling rate of the AWG is not violating this condition.

PR #646, opened 2021-03-12
Hot-fixing Keysight AWG

timoML:
1. ZeroDivisionError was raised for a debug message measuring upload times. If the upload fails time.time() may return idential values, causing the error.
2. Waveforms containing "_ch" in their name couldn't be deleted due to erroneous string handling.

## Description
- use better time.perf_counter()
- catch and ignore ZeroDivisionError
- use regex to split name strings

## How Has This Been Tested?
Tested on setup3

## Types of changes

## Checklist:


Neverhorst (approved):
Thanks

PR #650, opened 2021-04-07
Consistent naming of 'tau' in predefined methods

timoML:
This PR introduces a consistent naming of tau in all predefined methods in order to avoid confusion between the tau definitions.
We now have in code and the pulsed gui x-axis:
- tau: as defined by the filter function of a dd-like experiment
- tau_pulse_spacing: physical spacing between MW pulses.

This PR changes the experiments Ramsey and Hahn echo to use the first definition. Thus, compared to before, "tau" is now corrected for the pi pulse duration there. **End-users should be informed about this change** prior to merging.

In addition, new cw dd methods have been contributed by @NicolasStaudenmaier.

## Description

Provides a general function pulse_objects::tau_2_pulse_spacing() that should be used to transform from tau (filter function definition) to the pulse spacing between pulses. Right now, this function just substracts the duration of a pi pulse and does some sanity checking. It can be extended to different transformations by providing a function pointer.

In detail changes to the predefined methods:
- Rabi: gui x-axis "tau" -> "tau_pulse_spacing"
- T1: gui x-axis "tau" -> "tau_pulse_spacing"
- Ramsey: **now uses tau** (filter function definition)
- Hahn echo: **now uses tau** (filter function definition)
- XY8: only code cleanup. 
- Advanced DD with KDD, etc.: can be found in internal iqo-sequences repo. They need some new code in sampling_functions::DDMethods that is included in this PR. Contributed by @genko8. 
- Continuous DD: newly introduced

## Motivation and Context
Avoid confusion between different definitions.

## How Has This Been Tested?
In dummy mode, all changed functions have been queried with tau values that might be problematic. Warning work as expected.
@anjusha-vs is currently testing the advanced dd sequences (iqo-sequences repo), but they are not relevant for this PR.

## Types of changes

## Checklist:


Neverhorst:
If I'm not mistaken, you now label the x-axis with something like "tau_spacing" but provide the x-values for the "ideal" tau, i.e. the DD-filter-tau.
This is certainly not intended and as far as I recall we decided to display and save to data the "ideal" tau (the one used for post-processing in most sequences).

timoML:
Could you point to an exemplary sequence? That's definitely not intended. The tau_pulse_spacing label should only be used when there is no meaningful filter function (eg. T1).

Neverhorst:
> Could you point to an exemplary sequence? That's definitely not intended. The tau_pulse_spacing label should only be used when there is no meaningful filter function (eg. T1).

You are right, I accidentally jumped generate methods while reading. My bad...

logic/pulsed/pulse_objects.py
```diff
@@ -1099,6 +1100,51 @@ def sample_rate(self):
...
+                           custom_func=[None, None], **custom_kwwargs):
+        """
+        Converts tau to the physical pulse spacing between (microwave) pulses.
+        By definition, tau = 1/f where f is the filter frequency of a dynamical decoupling
+        experiment. For many cases this tau equals the time between the center of
+        consecutive pi pulses.
+        Thus, the default behavior is to subtract the duration of a pi pulse from tau.
+
+        :param t: tau (or tau_phys, if inverse==True) to be converted.
+        :param bool inverse: do the inverse transformation tau -> tau_phys
```
timoML, line 1114: Tau_phys is a left-over, should be replaced with tau_pulse_spacing

Neverhorst:
From a code point-of-view this looks good. It's a bit frustrating that git makes it so hard to refactor code while keeping attribution (blame).

Neverhorst (approved):
Great! Thanks everyone.

PR #652, opened 2021-05-11
Allow data to be saved in subfolders

timoML:
Currently, calling save_data() to a DataStorage object will always save to a constant directory that is fixed during construction of the DataStorage object.
This PR introduces a lightweight possibility to specify a sub directory.
It might come in handy when an external script is running the same measurement on different POI targets (see screenshot).

## Description
To be minimal code invasive, the nametag is checked for a possibly containing path. Eg. instead of using a 'pulsed_measurement' nametag, one could specify 'nv0/pulsed_measurement'. Although it overloads the nametag with additional functionality, I find this more elegant than introducing new parameters in all methods of the save toolchain. Importantly, external scripts can still use the save_measurement_data() methods of any module and would just need to alter the nametag provided.

## How Has This Been Tested?
Tested on dummy and saving pulsed measurement data. As saving is not implemented for most modules yet, it's currently hard to expand testing.

image

## Types of changes

## Checklist:

### DIFF
--- a/qudi/util/datastorage.py
+++ b/qudi/util/datastorage.py
@@ -175,11 +175,21 @@ class DataStorageBase(metaclass=ABCMeta):
 
         @param datetime.datetime timestamp: optional, timestamp to construct a generic filename from
         @param str filename: optional, filename to use (nametag and timestamp will be ignored)
-        @param str nametag: optional, nametag to include in the generic filename
+        @param str nametag: optional, nametag to include in the generic filename.
+                            May include sub directory, eg. "<subdir>/<nametag>".
         @param str file_extension: optional, the file extension to use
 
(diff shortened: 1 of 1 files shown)


timoML:
I allowed myself to add a fix for the NpyDataStorage. Please have a look whether this is the intended behavior. kwargs['comments'] is definetly not implemented in DataStorageBase.

alrik-durand:
I don't have any comment on this feature itself but I would like you mention that additional parameters to save logic can be used to easily sort data by POI or any other parameters (multiple parameters if you like !) Thus removing the need to create specific folders.
We do this in our lab every day and it has the advantage of no breaking the usual file structure :)

PR #653, opened 2021-05-11
New qudi IPython kernel concept

Neverhorst:
## Description
New implementation of the qudi IPython kernel that contains the module namespace of a running qudi instance.


kay-jahnke (approved):
Tested on Win 10 20H2 with dummy config of the new core.

All looks very good now. Thanks so much!

PR #655, opened 2021-05-12
[new-core] Comments in saved data

timoML:
Allow a comment string to be passed to a data storage object.

## Description
DataStorageBase defines a comment parameters in save_data(). For text, csv, npy I implemented the handling of the passed string to a comment section in the saved file.

## How Has This Been Tested?
Tested in newcore dummy for pulsed gui with .npy, .txt, .csv files.

## Screenshots 
image

## Types of changes

## Checklist:

### DIFF
--- a/qudi/util/datastorage.py
+++ b/qudi/util/datastorage.py
@@ -361,7 +361,7 @@ class TextDataStorage(DataStorageBase):
 
         if all_metadata:
             header_lines.append('Metadata:')
-            header_lines.append('===========')
+            header_lines.append('=========')
             for param, value in all_metadata.items():
                 if isinstance(value, (float, np.floating)):
                     header_lines.append(f'{param}: {value:.18e}')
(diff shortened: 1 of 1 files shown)


alrik-durand:
Thanks for this contribution.

If this change is merged, it will break the code we are using to parse the metadata automatically.
This is something we can live with, as this is the price of living a changing world.
But if we are starting from a clean slate, could I suggest using a fixed convention to mark the beginning and end of the metadata ?
I'm thinking of something like : 
```
#Metadata:
#===========
#---
#temperature: 4
#some other key: 1
#---
#Data:
#=====
#Signal(counts)
```

Here a line (`#---`) delimits the beginning and the end of the metadata section to parse. 
Maybe it should be `##` at the beginning of each line instead ?
Anyway, using a clear convention could help keep backward compatibility from now on.

Any thoughts on this ?

Neverhorst:
@alrik-durand As soon as the new core goes live, you will not need your custom parser anymore. Each data storage object (`TextDataStorage`, `NpyDataStorage` etc.) not only has a `save_data` method but also a `load_data` method (not yet implemented) that returns the data array as well as the metadata.  
Each storage object must be able to read back what it has saved. And since it is a standalone tool now, you can also import and use this funtionality without a running qudi instance.

If this is by any reason not an option, it is easy to write a parser that just relies on reading a consecutive block of key-value pairs following:
```
# Metadata:
# =======
```
until an empty line is encountered or you reach
```
# Data:
# ====
```

Neverhorst:
I have already manually merged this branch now while working on the data storage concept. closing it therefore. Sorry for not going through this PR.

PR #666, opened 2021-07-12
[core_pyside2] enhance system tray + extend timeouts

kay-jahnke:
- The system tray was not functional because the signals were not connected. Fixed that.
- Also added all GUI modules to the system tray.
- Extended the timeout of rpyc to 1 hour to allow long running cells in a notebook.
- Extended the timeout to declare a kernel dead to 10 seconds to allow for slower PCs that need long to start a new kernel.

## Motivation and Context

Now by selecting the GUI in the system tray it can be brought to the front, even if the window was closed.

## How Has This Been Tested?

Tested with dummy on core_pyside2 for Windows 10 21H1.

grafik

## Types of changes

## Checklist:

### DIFF
--- a/qudi/core/gui/gui.py
+++ b/qudi/core/gui/gui.py
@@ -58,10 +58,15 @@ class SystemTrayIcon(QtWidgets.QSystemTrayIcon):
         self.exiticon.addFile(os.path.join(iconpath, 'application-exit.png'), QtCore.QSize(16, 16))
         self.quitAction = QtWidgets.QAction(self.exiticon, 'Quit', self.right_menu)
 
+        self.restarticon = QtGui.QIcon()
+        self.restarticon.addFile(os.path.join(iconpath, 'view-refresh.png'), QtCore.QSize(16, 16))
+        self.restartAction = QtWidgets.QAction(self.restarticon, 'Restart', self.right_menu)
+
         self.left_menu.addAction(self.managerAction)
         self.left_menu.addSeparator()
 
         self.right_menu.addAction(self.quitAction)
+        self.right_menu.addAction(self.restartAction)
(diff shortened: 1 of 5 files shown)


qudi/core/gui/gui.py
```diff
@@ -245,6 +289,7 @@ def close_system_tray_icon(self):
         """
         self.hide_system_tray_icon()
         self.system_tray_icon.quitAction.triggered.disconnect()
+        self.system_tray_icon.restartAction.triggered.disconnect()
```
Neverhorst, line 292: Isn't 1h a bit excessive for a sync request timeout? I found little documentation for this parameter but to my understanding this does not mean the maximum time given to execute remote code (e.g. calling a remote function).

Neverhorst (commented):
Thank you.

Looks good in general. However I removed the explicit calls to the `Gui` singleton from `ModuleManager`. The manager should not care about a (optional) system tray. Connected the system tray to the `ModuleManager.sigModuleStateChanged` signal instead.
A good rule of thumb is: "If you need the reference to qudi main application for your new feature, there is most probably a better way."

Could you review these changes as well @kay-jahnke to see if everything is still working as you intended?

kay-jahnke:
tested it again at my end, and your changes work fine.

qudi/core/gui/gui.py
kay-jahnke: We had encountered problem for long running jupyter notebooks: they do through the sync error after a while, even though the code was still running in the background. So we solved this by increasing the timeout.

The more elegant solution would be to let rpyc know that something is still running (e.g. by a heartbeat) but I don't know how to do that.

Neverhorst (approved):

PR #670, opened 2021-07-22
[core_pyside2] startup fix

kay-jahnke:
## Description

Actually starting the modules listed under "startup" in the config.

## Motivation and Context

The startup parameter was loaded in the config, but the modules were not started.

## How Has This Been Tested?

Tested with custom config on Windows 10 21H1 for the following cases: 
- missing startup
- empty startup
- valid startup 
- startup with incorrect module name

## Types of changes

## Checklist:

### DIFF
--- a/qudi/core/application.py
+++ b/qudi/core/application.py
@@ -218,6 +218,12 @@ class Qudi(QtCore.QObject):
         self.gui = Gui(qudi_instance=self, stylesheet_path=self.configuration.stylesheet)
         self.gui.activate_main_gui()
 
+    def _start_startup_modules(self):
+        for module in self.configuration.startup_modules:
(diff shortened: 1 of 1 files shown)


Neverhorst:
I mean in principle the code looks good.
I'm unable to test it right now with an actual setup.
However, I'm a but concerned about loading a huge number of complex modules before starting the Qt event loop. It's probably irrelevant but one should test this with complex modules to be sure.

kay-jahnke:
I actually had already though about the starting of the modules before the event loop, but I did not find a better place for it.
Is there a hook which we could use that let's us start it after the event loop?

Neverhorst (approved):
Smileyface

PR #673, opened 2021-07-30
[core_pyside2] fix uic stream

kay-jahnke:
The new loadUi throws a ResourceWarning on python 3.8.

## Description

The ResourceWarning comes from an open IOWrapper around a subprocess which compiles the .ui file into python qt objects. Just closing the stream after it has been read solves the problem.

## Motivation and Context

Bug fix for a warning that appears under python 3.8 but does not seem to show up under 3.7.

## How Has This Been Tested?

Tested on hardware with loading a .ui file in an python 3.8 environment.

## Types of changes

## Checklist:

### DIFF
--- a/qudi/core/gui/uic.py
+++ b/qudi/core/gui/uic.py
@@ -29,6 +29,7 @@ __all__ = ('loadUi',)
 import os
 import re
 import tempfile
+import subprocess
 from importlib.util import spec_from_loader, module_from_spec
 from qudi.core.paths import get_artwork_dir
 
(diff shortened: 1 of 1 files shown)


Neverhorst (commented):
Good point @kay-jahnke. Actually using `os.popen()` is not recommended anyways, so I switched to `subprocess.run()`. This is more robust than closing the stream manually. It also checks the return value of the subprocess and raises an exception if needed. 

Improved the exception handling a bit while I was at it.

Can you check again if it works for you?

kay-jahnke:
@Neverhorst tested your modifications again and they work as intended. THX

Neverhorst (approved):

PR #699, opened 2022-04-15
Metadata GUI update

alrik-durand:
## Description
I've added pyperclip to the env files and added the missing docstrings


alrik-durand:
My mistake, wrong repo...

PR #704, opened 2022-08-09
Fixed two typing errors, where float values are passed to PyQt5 metho…

sidneycadot:
Fixed two typing errors, where float values are passed to PyQt5 methods that expect integers.

This issue leads to exceptions being thrown in (at least) Python 3.10.6 with PyQt5.

The solution is to round the float values prior to passing them as parameters.

it is conceivable that more issues like this are present in the code-base, but I haven't encountered them. Perhaps a thorough use of mypy could help to spot them.


Neverhorst (approved):
Thank you @sidneycadot

PR #708, opened 2022-12-06
Handle the case of an ROI with only one POI.

nbeaver:
This allows the POI manager to load an ROI file with a single POI.

## Description

Add the `ndmin` argument to `loadtxt` to ensure the shape is correct even when there is only one POI to load. Previously the resulting shape for `poi_names` was `()` and for `poi_coords` was `(3,)`, but with the fix it now has the correct shapes `(1,)` and `(1, 3)`.

## Motivation and Context

Currently, the POI manager will fail with an IndexError when attempting to load an ROI with a single ROI since [`loadtxt`](https://numpy.org/doc/stable/reference/generated/numpy.loadtxt.html) [defaults to](https://stackoverflow.com/questions/13528053/numpy-loadtxt-single-line-row-as-list) [squeezing the dimensions](https://stackoverflow.com/questions/23893625/numpy-loadtxt-for-one-and-more-input-lines).

## How Has This Been Tested?

I have tested it manually by loading an ROI with a single POI.

## Types of changes

## Checklist:

### DIFF
--- a/logic/poi_manager_logic.py
+++ b/logic/poi_manager_logic.py
@@ -1097,11 +1097,11 @@ class PoiManagerLogic(GenericLogic):
             filetag = filename.rsplit('_poi_list.dat', 1)[0]
 
         # Read POI data as well as roi metadata from textfile
-        poi_names = np.loadtxt(complete_path, delimiter='\t', usecols=0, dtype=str)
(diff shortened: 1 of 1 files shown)


timoML:
I can confirm the bug and the fix looks fine.
@nbeaver Do you want to contribute this to qudi-iqo-modules as an PR as well? Alternatively, I can also push your fix there but this would drop your credit.

